PR #97, opened 2014-08-28
[WIP] Workflow build step

tfennelly:
Changes to allow the plugin be used in a Workflow.  Would be good if @slide (et al) and @jglick could review.  In particular... look for `TODO` markers in the changes I made.  These were things I was not sure about.

Atm it works using the base 'step' step (i.e. you need to specify the $class) e.g.
```
node('master') {
	catchError {
		// Do stuff ....
	}
	step($class: 'hudson.plugins.emailext.WorkflowPublisher', recipientList: 'tfennelly@localhost', subject: 'Build failed', content: 'The build failed', attachBuildlog: true)
}
```

I pulled the workflow specific code up into an extension of `ExtendedEmailPublisher`.  I think it's cleaner this way.

Once the workflow apis are published (?) we can build a dedicated step for this, it would look something like the following (see [BuildStep-Job-with-step](https://github.com/tfennelly/email-ext-plugin/tree/BuildStep-Job-with-step))
```
node('master') {
	catchError {
		// Do stuff ....
	}
	emailext(recipientList: 'tfennelly@localhost', subject: 'Build failed', content: 'The build failed', attachBuildlog: true)
}
```


jglick:
Given https://github.com/jenkinsci/workflow-plugin/pull/39 I wonder if this should just be closed: it is more straightforward to just use a plain `mail` step with some inline Groovy templating then configure a very complex step in the Snippet Generator UI and paste it in. There would still be some reasons some form of integration with this plugin could be fruitful:

* If some of the specialized kinds of content (mainly for the message body I think) require complex calculations that it would be unpleasant to write from scratch. For example, `WorkspaceFileContent` adds little value to a workflow, and `ScriptContent` is out of the question, but `BuildLogRegexContent` is doing real work that could not easily be replicated.
* If some of the triggers are not trivially written inline in the flow. (Assuming that the current `WorkflowRun` and relevant methods were made accessible: [discussion](https://github.com/jenkinsci/junit-plugin/pull/9/files#r22653338).) `ImprovementTrigger` is a good example.
* Custom `RecipientProvider`s, assuming some API changes in core to make it possible to use them on non-`AbstractBuild`s.

In other words, rather than offering `ExtendedEmailPublisher` as a monolithic build step, offer individual steps which wrap its useful extension points à la carte, so that a workflow could be written which (after `catchError`) assembles some canned functionality along with short bits of customized logic and finally calls `mail` with the result.

s0undt3ch:
:+1: 

Recreating the email body logic is possible, but preferably done outside of the workflow script

AnalogJ:
Sorry to dig this up again. After looking at this PR and some of the other PR's in the workflow plugin I'm a bit confused. I'm trying to figure out if the email-ext plugin or the `mail` step in the workflow plugin supports email attachments yet. 

Thanks

slide:
I've never used the workflow plugin, so I don't know how it works with email-ext. Email-ext itself supports attachments, but I don't know how that is exposed in workflow, if at all.

jglick:
@AnalogJ better to use a different forum such as Stack Overflow for questions, or just use JIRA. (In this case, no, there is no attachment support yet.)

AnalogJ:
@jglick @slide  Thanks for following up. The only reason I posted here is that the workflow-plugin compatibility page specifically [links](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md) to this PR. I'll make a new question on SO for a mail attachment workaround.

Thanks!

vbansal22:
Sorry to stir the pot again. Attachment support is a much wanted feature in the workflow mail step. Having just "Please go to ${env.BUILD_URL}." in the body is not sufficient because the recipient could essentially be viewing the message on a mobile phone and may not have access to the build machine at that time. Is there a way to read the build log from somewhere in the script itself?

vbansal22:
Getting the build output in the email is not such a big issue. I could do it this way.
    try {
        ...
   } catch (e) {
        sh "curl ${env.BUILD_URL}/consoleText -o buildoutput"
        def buildoutput = readFile('buildoutput')

        mail (to: 'abc@example.com',
        from: 'build-bot-donot-reply@tibco.com',
        subject: "Job '${env.JOB_NAME}' (${env.BUILD_NUMBER}) failed ",
        body: "Please go to ${env.BUILD_URL}. or see as follows: ${buildoutput}");
        throw e
}

jglick:
@vbansal22 probably what you are looking for is [JENKINS-28119](https://issues.jenkins-ci.org/browse/JENKINS-28119).

PR #99, opened 2014-10-17
Enhance groovy-html.template

KostyaSha:


slide:
Should this still be open? Is it still being worked on?

KostyaSha:
I switched to other issues. But i think this changes are still valid.

slide:
So, it's not a work in progress anymore?

KostyaSha:
@slide i think this portion of changes was completed. I also did some other improvements but forgot send. 
If you are ok with this changes, then please merge.

PR #100, opened 2015-01-08
Add uberClassLoader to JellyContext

Everspace:
Allows for a jelly email create classes from plugins, like [BuildFailureAnalyzer's ScanOnDemandTask](https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/main/java/com/sonyericsson/jenkins/plugins/bfa/sod/ScanOnDemandTask.java).

This is already done on the groovy-script side, so I've added it here.


slide:
Did you run the test suite and there were no issues?

PR #101, opened 2015-01-21
Add script content tag

Everspace:
I really want to be able to include various scripts to make a *build your own* approach. The `${FILE}` tag is work-space bound, and this is a variation on it that is server-side. It externalizes script content without having to resort to scriptler/groovy shenanigans which also opens up version control options.

    ${SCRIPT_CONTENT,script="email-templates/scripts/helloWorld.groovy"}

produces:

    msg.setSubject("Hello from presend-script land!")


src/main/java/hudson/plugins/emailext/plugins/content/UnevaluatedScriptContent.java
```diff
@@ -0,0 +1,74 @@
...
+
+/*
+    @Override
+    evaluate(AbstractBuild<?, ?> context, TaskListener listener, String macroName, 
+            Map<String, String> arguments, ListMultimap<String, String> argumentMultimap) 
+            throws MacroEvaluationException, IOException, InterruptedException {
+
+    }
+*/
+    @Override
```
slide: Please get rid of commented code.

slide:
It looks like your tree is messed up or something, this merge request is pulling in commits that shouldn't be in it.

Everspace:
I have no idea what I'm doing with git. I'll try again.

PR #103, opened 2015-02-24
Fix multibyte text corruption in email sender name

orih:
If the `System Admin e-mail address` contains multibyte characters,

jenkins-configuration

the sender text received is corrupts.

recieved-email-header

I confirmed this problem on

- Jenkins 1.585
- email-ext-plugin 2.39

and

- Jenkins 1.580.1 (development environment)
- email-ext-plugin 04b383 (master branch)

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -528,25 +528,6 @@ public class ExtendedEmailPublisher extends Notifier implements MatrixAggregatab
 
         MimeMessage msg;
 
-        // If not overriding global settings, use the Mailer class to create a session and set the from address
-        // Else we'll do it ourselves
-        Session session;
-        if (!overrideGlobalSettings) {
-            debug(context.getListener().getLogger(), "NOT overriding default server settings, using Mailer to create session");
-            session = Mailer.descriptor().createSession();
-            msg = new MimeMessage(session);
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
```diff
@@ -535,27 +544,18 @@ private MimeMessage createMail(ExtendedEmailPublisherContext context) throws Mes
             debug(context.getListener().getLogger(), "NOT overriding default server settings, using Mailer to create session");
             session = Mailer.descriptor().createSession();
             msg = new MimeMessage(session);
-            msg.setFrom(new InternetAddress(JenkinsLocationConfiguration.get().getAdminAddress()));
+            msg.setFrom(Mailer.StringToAddress(JenkinsLocationConfiguration.get().getAdminAddress(), charset));
```
slide: I'd rather not introduce another dependency on the Mailer code here.

orih: Then I can re-implement a method like `Mailer.StringToAddress` as a private method. I think it's not good way. How do you think?

slide:
I'll fix this a different way.

orih:
Okay then let me know if you need some multibyte data or to test your fixed code :)

slide:
Fixed in 62e9ad4a5a51d3bbf765d78d49438848a220d426

orih:
Thanks. I'll try it later.

orih:
It works perfect! Thank you so much.

PR #107, opened 2015-09-19
Add workflow step. Update to using Run<?,?> instead of AbstractBuild<…

slide:
Add a workflow step and update many places to use Run instead of AbstractBuild


slide:
@jglick Can you please review the workflow stuff on this?

jglick:
Supersedes #97?

slide:
Yes, it supercedes #97

src/main/java/hudson/plugins/emailext/EmailExtStep.java
```diff
@@ -0,0 +1,191 @@
...
+                publisher.attachmentsPattern = step.attachmentsPattern;
+            }
+
+            if(StringUtils.isNotBlank(step.mimeType)) {
+                publisher.contentType = step.mimeType;
+            }
+
+            final ExtendedEmailPublisherContext ctx = new ExtendedEmailPublisherContext(publisher, run, workspace, launcher, listener);
+            final Multimap<String, EmailTrigger> triggered = ArrayListMultimap.create();
+            triggered.put(AlwaysTrigger.TRIGGER_NAME, publisher.configuredTriggers.get(0));
```
jglick: Is `triggered` being used for anything?

src/main/java/hudson/plugins/emailext/EmailExtStep.java
```diff
@@ -0,0 +1,191 @@
...
+            publisher.sendMail(ctx);
+            return null;
+        }
+    }
+
+
+    @Extension
+    public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+
+        public static final String defaultMimeType = "text/plain";
```
jglick: Looks like you [forgot to use this](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md#handling-default-values)?

jglick:
I still think my (vague) [earlier proposal](https://github.com/jenkinsci/email-ext-plugin/pull/97#issuecomment-71212666) could be useful in the long run, but in the meantime this looks like it brings much-needed basic support, so :+1: overall. I am still sifting through the details.

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
```diff
@@ -34,16 +36,18 @@ public CulpritsRecipientProvider() {
...
-        for (User user : users) {
-            if (!EmailRecipientUtils.isExcludedRecipient(user, context.getListener())) {
-                String userAddress = EmailRecipientUtils.getUserConfiguredEmail(user);
-                if (userAddress != null) {
-                    descriptor.debug(context.getListener().getLogger(), "Adding user address %s, they were not considered an excluded committer", userAddress);
-                    EmailRecipientUtils.addAddressesFromRecipientList(to, cc, bcc, userAddress, env, context.getListener());
-                } else {
-                    context.getListener().getLogger().println("Failed to send e-mail to " + user.getFullName() + " because no e-mail address is known, and no default e-mail domain is configured");
+        Run<?,?> run = context.getRun();
+        if(run instanceof AbstractBuild) {
```
recena: `if (` space

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
```diff
@@ -456,6 +456,7 @@ private boolean executePresendScript(ExtendedEmailPublisherContext context, Mime
 
             Binding binding = new Binding();
             binding.setVariable("build", context.getBuild());
+            binding.setVariable("run", context.getRun());
```
jglick: Is it really needed? The term “build” is widely accepted, including for a `WorkflowRun`. Anyway there is no need for a `presendScript` for Workflow.

src/main/java/hudson/plugins/emailext/plugins/content/AbstractChangesSinceContent.java
```diff
@@ -115,13 +113,13 @@ public boolean printSpec(StringBuffer buf, char formatChar) {
                 switch (formatChar) {
                     case 'c':
                         try {
-                            buf.append(changes.evaluate(currentBuild, listener, ChangesSinceLastBuildContent.MACRO_NAME));
+                            buf.append(changes.evaluate((AbstractBuild)currentRun, listener, ChangesSinceLastBuildContent.MACRO_NAME));
```
jglick: Unsafe cast.

src/main/java/hudson/plugins/emailext/plugins/content/ChangesSinceLastBuildContent.java
```diff
@@ -85,9 +86,9 @@ public String evaluate(AbstractBuild<?, ?> build, TaskListener listener, String
             buf.append(def);
         }
         if (showDependencies) {
-            AbstractBuild previousBuild = ExtendedEmailPublisher.getPreviousBuild(build, listener);
-            if (previousBuild != null) {
-                for (Entry<AbstractProject, DependencyChange> e : build.getDependencyChanges(previousBuild).entrySet()) {
+            Run<?,?> previousRun = ExtendedEmailPublisher.getPreviousRun(build, listener);
+            if (previousRun != null && previousRun instanceof AbstractBuild) {
```
jglick: BTW `instanceof` subsumes a null check.

src/main/java/hudson/plugins/emailext/plugins/recipients/DevelopersRecipientProvider.java
```diff
@@ -29,8 +30,12 @@ public DevelopersRecipientProvider() {
     public void addRecipients(ExtendedEmailPublisherContext context, EnvVars env, Set<InternetAddress> to, Set<InternetAddress> cc, Set<InternetAddress> bcc) {
         ExtendedEmailPublisherDescriptor descriptor = Jenkins.getInstance().getDescriptorByType(ExtendedEmailPublisherDescriptor.class);
         Set<User> users = new HashSet<User>();
-        for (ChangeLogSet.Entry change : context.getBuild().getChangeSet()) {
-            users.add(change.getAuthor());
+
+        if (context.getRun() instanceof AbstractBuild) {
+            AbstractBuild<?, ?> build = (AbstractBuild) context.getRun();
+            for (ChangeLogSet.Entry change : build.getChangeSet()) {
```
jglick: You can probably use `getChangeSets` via reflection, pending [JENKINS-24141](https://issues.jenkins-ci.org/browse/JENKINS-24141).

src/main/resources/hudson/plugins/emailext/EmailExtStep/config.jelly
```diff
@@ -0,0 +1,50 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="to" title="To">
+        <f:textbox />
```
jglick: This is optional according to the step definition, so it would be good to make that clear (for example in `help-to.html`).

pom.xml
```diff
@@ -144,6 +145,34 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency> <!-- StepConfigTester -->
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-step-api</artifactId>
+            <classifier>tests</classifier>
+            <version>${workflow.version}</version>
+            <scope>test</scope>
+        </dependency>
+
```
recena: I think It is enough (for testing) with...
```
<dependency>
    <groupId>org.jenkins-ci.plugins.workflow</groupId>
    <artifactId>workflow-aggregator</artifactId>
    <version>${workflow-jenkins-plugin.version}</version>
    <scope>test</scope>
</dependency>
```

src/test/java/hudson/plugins/emailext/EmailExtStepTest.java
```diff
@@ -0,0 +1,28 @@
...
+import org.jenkinsci.plugins.workflow.steps.StepConfigTester;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.recipes.WithPlugin;
+
+/**
+ * Created by acearl on 9/15/2015.
+ */
+public class EmailExtStepTest {
```
jglick: Good idea to add a test actually running the step once, using `mock_javamail` I suppose.

src/main/java/hudson/plugins/emailext/AttachmentUtils.java
```diff
@@ -222,13 +222,13 @@ private static void attachSingleLog(ExtendedEmailPublisherContext context, Run<?
     }
     
     public static void attachBuildLog(ExtendedEmailPublisherContext context, Multipart multipart, boolean compress) {
-        if(context.getBuild() instanceof MatrixBuild) {
-            MatrixBuild build = (MatrixBuild)context.getBuild();
+        if(context.getRun() instanceof MatrixBuild) {
```
recena: `if (`space

src/main/java/hudson/plugins/emailext/EmailExtStep.java
```diff
@@ -0,0 +1,191 @@
...
+
+    private boolean compressLog;
+
+    @DataBoundConstructor
+    public EmailExtStep(String subject, String body) {
+        this.subject = subject;
+        this.body = body;
+    }
+
+    public @CheckForNull String getAttachmentsPattern() {
```
recena: Can it be `null`?

src/main/java/hudson/plugins/emailext/EmailExtStep.java
```diff
@@ -0,0 +1,191 @@
...
+        return attachmentsPattern == null ? "" : attachmentsPattern;
```
recena: Can it be `null`? In `setAttachmentsPattern` method you are validating the input.

pom.xml
jglick: Well this PR is using `StepConfigTester`. The `org.jenkins-ci.plugins.workflow:workflow-aggregator:*:tests` dependency appears unused for now.

src/main/java/hudson/plugins/emailext/EmailExtStep.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @DataBoundSetter
+    public void setAttachmentsPattern(@CheckForNull String attachmentsPattern) {
+        if(StringUtils.isNotBlank(attachmentsPattern)) {
```
recena: `if (` space

slide:
@jglick The main thing I have seen people ask for is attachments. This is one of the reasons that I went down this route.

src/main/java/hudson/plugins/emailext/EmailExtStep.java
```diff
@@ -0,0 +1,191 @@
...
+
+        @StepContextParameter
+        private transient FilePath workspace;
+
+        @Override
+        protected Void run() throws Exception {
+            ExtendedEmailPublisher publisher = new ExtendedEmailPublisher();
+            publisher.configuredTriggers.clear();
+
+            AlwaysTrigger.DescriptorImpl descriptor = Jenkins.getInstance().getDescriptorByType(AlwaysTrigger.DescriptorImpl.class);
```
amuniz: `Jenkins.getActiveInstance()` to avoid NPE (although Jenkins instance must be always available when running a step).

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherContext.java
```diff
@@ -12,15 +15,26 @@
...
-    
+
+    @Deprecated
     public ExtendedEmailPublisherContext(ExtendedEmailPublisher publisher, AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) {
         this.publisher = publisher;
-        this.build = build;
+        this.run = build;
+        this.workspace = build.getWorkspace();
+        this.launcher = launcher;
+        this.listener = listener;
```
amuniz: Wouldn't be better to directly call the new constructor? Just for easier maintenance in the future.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
```diff
@@ -748,20 +749,20 @@ public BuildStepMonitor getRequiredMonitorService() {
...
      * actual previous build is still in progress
      * @return the previous build, or null if that build is missing, or is still
      * in progress
      */
     public static @CheckForNull
-    AbstractBuild<?, ?> getPreviousBuild(@Nonnull AbstractBuild<?, ?> build, TaskListener listener) {
-        AbstractBuild<?, ?> previousBuild = build.getPreviousBuild();
-        if (previousBuild != null && previousBuild.isBuilding()) {
-            listener.getLogger().println(Messages.ExtendedEmailPublisher__is_still_in_progress_ignoring_for_purpo(previousBuild.getDisplayName()));
+    Run<?, ?> getPreviousRun(@Nonnull Run<?, ?> run, TaskListener listener) {
```
amuniz: I don't know if any other plugin (that depends on this one) is using this public method. If so, this change will break binary compatibility. If you want to fix it you can use `@WithBridgeMethods` ([in this way](https://github.com/jenkinsci/analysis-core-plugin/blob/42c50ab8f5cca9da0d83519a03afd7fe2a46de57/src/main/java/hudson/plugins/analysis/views/ConsoleDetail.java#L114))

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherContext.java
```diff
@@ -32,15 +46,23 @@ public ExtendedEmailPublisher getPublisher() {
     protected void setPublisher(ExtendedEmailPublisher publisher) {
         this.publisher = publisher;
     }
-    
+
+    @Deprecated
```
amuniz: When developing, a Javadoc reference to the new method is really useful.

src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
```diff
@@ -68,7 +68,9 @@ public static String transformText(String origText, ExtendedEmailPublisherContex
             List<TokenMacro> macros = new ArrayList<TokenMacro>(getPrivateMacros());
             if(additionalMacros != null)
                 macros.addAll(additionalMacros);
-            newText = TokenMacro.expandAll(context.getBuild(), context.getListener(), newText, false, macros);
+            if(context.getBuild() != null) {
+                newText = TokenMacro.expandAll(context.getBuild(), context.getListener(), newText, false, macros);
```
amuniz: Does not this apply to when using it inside a Workflow build?

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
```diff
@@ -34,16 +36,18 @@ public CulpritsRecipientProvider() {
...
+            Set<User> users = ((AbstractBuild<?,?>)run).getCulprits();
+            for (User user : users) {
+                if (!EmailRecipientUtils.isExcludedRecipient(user, context.getListener())) {
+                    String userAddress = EmailRecipientUtils.getUserConfiguredEmail(user);
+                    if (userAddress != null) {
+                        descriptor.debug(context.getListener().getLogger(), "Adding user address %s, they were not considered an excluded committer", userAddress);
+                        EmailRecipientUtils.addAddressesFromRecipientList(to, cc, bcc, userAddress, env, context.getListener());
+                    } else {
+                        context.getListener().getLogger().println("Failed to send e-mail to " + user.getFullName() + " because no e-mail address is known, and no default e-mail domain is configured");
+                    }
```
amuniz: Is all this code supposed to not run when `run instanceof WorkflowRun == true`?

src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
KostyaSha: As i understand that wouldn't work and i also depend on `TokenMacro.expandAll(build, ...` method. First of all TokenMacro + workflow should be resolved imho.

src/main/java/hudson/plugins/emailext/plugins/content/ChangesSinceLastBuildContent.java
KostyaSha: +1 null check with instanceof not required (you can run PMD to get such hints).

src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
slide: That's a much more difficult problem...

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
slide: I agree, I did a quick check and didn't see any other plugin using it. I felt ok changing it.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherContext.java
slide: Yes, it would, will update.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherContext.java
slide: Will do.

src/main/java/hudson/plugins/emailext/plugins/content/AbstractChangesSinceContent.java
slide: Will fix.

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
slide: Correct, currently I don't know of a way to get culprits when run is a WorkflowRun.

src/test/java/hudson/plugins/emailext/EmailExtStepTest.java
slide: Right, I'll do something similar to what is done in other areas of email-ext.

jglick:
I guess this has been merged?

slide:
Yes, it has been merged, I am cleaning it up a bit though.

PR #108, opened 2015-10-09
Support AWS SES Message-ID rewriting

weisslj:
This ist only a proof-of-concept patch, pull request created by request in https://issues.jenkins-ci.org/browse/JENKINS-30854

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
 
     <artifactId>email-ext</artifactId>
     <packaging>hpi</packaging>
-    <version>2.41-ses</version>
+    <version>2.41-SNAPSHOT</version>
     <name>Email Extension Plugin</name>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Email-ext+plugin</url>
     <issueManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
 
     <artifactId>email-ext</artifactId>
     <packaging>hpi</packaging>
-    <version>2.41-SNAPSHOT</version>
+    <version>2.41-ses</version>
     <name>Email Extension Plugin</name>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Email-ext+plugin</url>
     <issueManagement>
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -357,6 +357,9 @@ public class ExtendedEmailPublisher extends Notifier implements MatrixAggregatab
                         try {
                             ExtendedEmailPublisherDescriptor descriptor = getDescriptor();
                             String smtpHost = descriptor.getSmtpServer();
+                            if (smtpHost == null) {
+                                smtpHost = "";
+                            }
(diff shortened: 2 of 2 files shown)


weisslj:
See answer in [JENKINS-30854](https://issues.jenkins-ci.org/browse/JENKINS-30854?focusedCommentId=238590&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-238590).

pom.xml
```diff
@@ -17,7 +17,7 @@
 
     <artifactId>email-ext</artifactId>
     <packaging>hpi</packaging>
-    <version>2.41-SNAPSHOT</version>
+    <version>2.41-ses</version>
```
slide: I can't accept this change at all.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
```diff
@@ -352,7 +355,35 @@ private boolean sendMail(ExtendedEmailPublisherContext context) {
...
+                                try {
+                                    transport.connect();
+                                    transport.sendMessage(msg, allRecipients);
+                                    if (transport instanceof SMTPTransport) {
+                                        String response = ((SMTPTransport)transport).getLastServerResponse();
+                                        String[] parts = response.trim().split(" +");
+                                        if (parts.length == 3 && parts[0].equals("250") && parts[1].equals("Ok")) {
+                                            String MessageID = "<" + parts[2] + "@" + region + ".amazonses.com>";
+                                            msg.setHeader("Message-ID", MessageID);
+                                        }
```
slide: You should be able to do this part in the pre-send script, why do you not think that is possible? You've basically added something before the call to Transport.send, which is what the pre-send script is for.

pom.xml
weisslj: Of course, I just send the pull request as requested in [JENKINS-30854](https://issues.jenkins-ci.org/browse/JENKINS-30854), for easier review.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
weisslj: It is not possible, because the new message id is only available after the email has been sent, not before.

If I put the code that sends the actual email in the pre-send script, the email will be sent twice, and if I set `cancel = true`, the message id is lost for the In-Reply-Header.

slide: Ah, I was misreading the diff. I just really don't want to incorporate something specific like this. If it could be reworked into something generic, I would. You can always build a local version and use it as you are now, but unless its made more generic, I don't think I can include it.

weisslj: Yes, suspected that it might be too specific to include it. Would you accept a patch for an additional "post-send" script where the result of the last SMTP answer is available?

slide: Yeah, I think I could accept something like that.

weisslj: Cool, thanks! I created a new pull request in #109 for the new post-send scripts.

weisslj:
Closed in favor of #109

PR #109, opened 2015-10-17
Add post-send script feature

weisslj:
For some tasks pre-send scripts are not sufficient, like rewriting the Message-ID based on the SMTP response for later In-Reply-To headers. This is needed when using Amazon Simple Email Service (AWS SES) for sending emails.

Other possible use cases are scripts that act upon failed sending attempts.

Here is a example post-send script that does the message id rewriting for AWS SES:

```groovy
import com.sun.mail.smtp.SMTPTransport;
import java.util.regex.Matcher;
import java.util.regex.Pattern;
import hudson.plugins.emailext.ExtendedEmailPublisherDescriptor;

String smtpHost = props.getProperty("mail.smtp.host", "");
String awsRegex = "^email-smtp\\.([a-z0-9-]+)\\.amazonaws\\.com\$";
Pattern p = Pattern.compile(awsRegex);
Matcher m = p.matcher(smtpHost);
if (m.matches()) {
    String region = m.group(1);
    if (transport instanceof SMTPTransport) {
        String response = ((SMTPTransport)transport).getLastServerResponse();
        String[] parts = response.trim().split(" +");
        if (parts.length == 3 && parts[0].equals("250") && parts[1].equals("Ok")) {
            String MessageID = "<" + parts[2] + "@" + region + ".amazonses.com>";
            msg.setHeader("Message-ID", MessageID);
        }
    }
}
```

References:
- https://issues.jenkins-ci.org/browse/JENKINS-30854
- Pull request #108
- http://docs.aws.amazon.com/ses/latest/DeveloperGuide/smtp-response-codes.html

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -54,6 +54,7 @@ import org.codehaus.groovy.control.customizers.ImportCustomizer;
 import org.jenkinsci.plugins.tokenmacro.TokenMacro;
 import org.kohsuke.groovy.sandbox.SandboxTransformer;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * {@link Publisher} that sends notification e-mail.
@@ -152,18 +153,18 @@ public class ExtendedEmailPublisher extends Notifier implements MatrixAggregatab
 
     @Deprecated
     public ExtendedEmailPublisher(String project_recipient_list, String project_content_type, String project_default_subject,
-            String project_default_content, String project_attachments, String project_presend_script, String project_postsend_script,
+            String project_default_content, String project_attachments, String project_presend_script,
             int project_attach_buildlog, String project_replyto, boolean project_save_output,
             List<EmailTrigger> project_triggers, MatrixTriggerMode matrixTriggerMode) {
 
         this(project_recipient_list, project_content_type, project_default_subject, project_default_content,
-                project_attachments, project_presend_script, project_postsend_script, project_attach_buildlog, project_replyto,
+                project_attachments, project_presend_script, project_attach_buildlog, project_replyto,
                 project_save_output, project_triggers, matrixTriggerMode, false, Collections.EMPTY_LIST);
     }
 
     @DataBoundConstructor
     public ExtendedEmailPublisher(String project_recipient_list, String project_content_type, String project_default_subject,
-            String project_default_content, String project_attachments, String project_presend_script, String project_postsend_script,
+            String project_default_content, String project_attachments, String project_presend_script,
             int project_attach_buildlog, String project_replyto, boolean project_save_output,
             List<EmailTrigger> project_triggers, MatrixTriggerMode matrixTriggerMode, boolean project_disabled,
             List<GroovyScriptPath> classpath) {
@@ -173,7 +174,6 @@ public class ExtendedEmailPublisher extends Notifier implements MatrixAggregatab
         this.defaultContent = project_default_content;
         this.attachmentsPattern = project_attachments;
         this.presendScript = project_presend_script;
-        this.postsendScript = project_postsend_script;
         this.attachBuildLog = project_attach_buildlog > 0;
         this.compressBuildLog = project_attach_buildlog > 1;
         this.replyTo = project_replyto;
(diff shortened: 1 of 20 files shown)


src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
```diff
@@ -487,6 +507,58 @@ private boolean executePresendScript(ExtendedEmailPublisherContext context, Mime
         return !cancel;
     }
 
+    private void executePostsendScript(ExtendedEmailPublisherContext context, MimeMessage msg, Session session, Transport transport)
```
slide: Could the executePresendScript just be updated to be more generic and be called executeScript?

weisslj: Yes, I can try. See my attempt in e0f48c1.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
```diff
@@ -147,18 +152,18 @@
...
-                project_attachments, project_presend_script, project_attach_buildlog, project_replyto,
+                project_attachments, project_presend_script, project_postsend_script, project_attach_buildlog, project_replyto,
                 project_save_output, project_triggers, matrixTriggerMode, false, Collections.EMPTY_LIST);
     }
 
     @DataBoundConstructor
     public ExtendedEmailPublisher(String project_recipient_list, String project_content_type, String project_default_subject,
-            String project_default_content, String project_attachments, String project_presend_script,
+            String project_default_content, String project_attachments, String project_presend_script, String project_postsend_script,
             int project_attach_buildlog, String project_replyto, boolean project_save_output,
```
slide: Can you change this to a @DataBoundSetter on the field instead? That was if anyone is using the current constructor, it won't break them.

weisslj: Yes, see my attempt in 51ab6c3.

weisslj:
Thanks a lot!

Kern3l:
Hey, could you push a new release with this pull request? We'd love this feature. Thank you in advance!

slide:
Yes, I will be doing one soon. I need to clean-up some other stuff first.

weisslj:
@Kern3l Just out of interest: What are you planning to do as post-send script?

Sorry for abusing the pull request, there is no private messaging on GitHub anymore.

Kern3l:
@slide 
Awesome, can't wait for it!

BTW I'd LOVE if I could set up default "Send to" per trigger. For us one default "Send to" for all triggers would do, but maybe you'd like to use the same code as in . We send notifications only to emails from recipient list (which is our use case from standard jenkins mailer) and someone in the future will not take a peek at existing configuration's advanced settings due to hurry or anything. In the result it would defeat mailer's purpose when job fails and nobody would get an email about that.


@weisslj 
To do exactly what you do - add proper Message-ID to emails delivered via AWS, so they would be grouped together as a conversation.

kramarz:
We have updated plugin and restarted jenkins and this feature does not work for me. I have set up DEFAULT_POSTSEND_SCRIPT in main jenkins settings as in this pull request. When I check configuration in my project I see Postsend script as empty. Changing it to $DEFAULT_PRESEND_SCRIPT and saving does not produce any error, but when I open configuration again it is still empty. Putting script directly into project configuration also does not save. I can save Presend script normally - at least it stays after reopening of project configuration.
Is it a bug or something is wrong with my configuration? We are using jenkins docker image taged as jenkins:1.642.1

slide:
Sounds like a bug

weisslj:
@kramarz 
It works for me, version 1.625.2. I will upgrade to the latest version and check again.

You need to set "Default Post-send Script" in the main settings to the script above, and "Post-send Script" in the job settings to "$DEFAULT_POSTSEND_SCRIPT".

weisslj:
@kramarz 
After upgrading to latest Jenkins (1.642.2) and email-ext (2.41.3) version I have the same bug. I was on a custom build version of email-ext before.

I tried to figure out what when wrong, but so far no luck. It could be related to the upgrade process.

weisslj:
@kramarz 
I opened [JENKINS-33205](https://issues.jenkins-ci.org/browse/JENKINS-33205) for this bug. There is a workaround: Directly edit config.xml and then reload the configuration from disk.

weisslj:
@kramarz
See #132 for a patch.

kramarz:
Thanks for quick reaction. I will use your workaround for now.

PR #110, opened 2015-11-04
[JENKINS-31380] Remove token overlapped build variable.

KostyaSha:
https://issues.jenkins-ci.org/browse/JENKINS-31380


KostyaSha:
Token supports options so `${BUILD_ID()}` is expected as token. Seems they can't be removed from plugin.

slide:
Not sure what you mean it supports options. There are no parameters defined for BUILD_ID.

slide:
Strange, this works fine for me. It's done via the missingMethod for the context of the Groovy shell.

KostyaSha:
@slide see test case. As i understand `${VAR()}` expected to be a token even `VAR` exists as standard var. I will think about using groovy templates in some other way.

slide:
@KostyaSha yes, I know. This is handled via EmailExtScript which is set as the base class for the script for the groovy. When a "method" in this case something like ${BUILD_ID()} is used, the methodMissing in EmailExtScript will be called and will look at ALL TokenMacro's to see if there is something that fits. This is working locally for me after removing BUILD_ID from email-ext.

slide:
@KostyaSha I was still running on a codebase that had BUILD_ID as a TokenMacro. When I removed it, I found the issue as you did, but I have a fix for it.

slide:
Will fix this a different way.

PR #131, opened 2016-02-29
Fix JENKINS-33205 (disappearing post-send script project setting)

weisslj:
The post-send script setting disappears when saving job configuration.

This fixes up 51ab6c3, the `@DataBoundSetter` approach does not work as intended, second deprecated constructor does the job.

References:
- https://issues.jenkins-ci.org/browse/JENKINS-33205
- Pull request #109


weisslj:
New one opened in #132

PR #132, opened 2016-02-29
Fix JENKINS-33205 (disappearing post-send script project setting)

weisslj:
The post-send script setting disappears when saving job configuration.

This fixes up 51ab6c3 and adds a test.

References:
- https://issues.jenkins-ci.org/browse/JENKINS-33205
- Pull request #109

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -206,8 +206,8 @@ public class ExtendedEmailPublisher extends Notifier implements MatrixAggregatab
     }
 
     @DataBoundSetter
-    public void setPostsendScript(String project_postsend_script) {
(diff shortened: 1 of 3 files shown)


slide:
I can't really merge this since the constructor could be used elsewhere. We need to determine why the databoundsetter is not working.

weisslj:
But unlike before 51ab6c3 the old constructor is not deleted, but copied with `@Deprecated`, so it still exists and can be used elsewhere. At least that is my understanding.

slide:
Looks to me like the DataBoundSetter is not working because the field in the form needs to be named presendScript instead of project_postsend_script. I would rather see if that worked than add a new deprecated constructor.

weisslj:
@slide 
That worked, thanks! New commit 1832ae0 pushed forcefully to this branch, which also includes a regression test. Can you merge this one?

PR #142, opened 2016-09-30
enable setter for smtpHost in descriptor

keinkonzept:


davidvanlaatum:
Is this fixing a bug or adding for feature?

keinkonzept:
It's adding for a new feature. Basic e-mail notifier has this setter, along with setDefaultSuffix. It's needed to configure plugin programatically without overhead of submitting whole configuration JSON.

sgtcoolguy:
It'd be nice to have setters for **all** the properties, so that I could programmatically set up the plugin using groovy scripts. I use chef to configure our Jenkins master. I can write out overridden plugin config XML files to disk, but that can be a pain when hey contain values needing to be encrypted - plus I seem t need to restart Jenkins to get the values reflected, a simple reload config from disk doesn't cut it.

But using groovy to manipulate the plugins in code does work without a restart, and avoids needing to jump through hoops to encrypt the secret values before sticking them into the xml file - it's just that not all plugins expose an API to be able to configure them outside the web UI.

davidvanlaatum:
Create a issue in Jira and I'll do it at some stage

sgtcoolguy:
Awesome, thanks: https://issues.jenkins-ci.org/browse/JENKINS-39147

PR #148, opened 2017-01-06
change ListRecipientProvider for pipeline code

qxo:
change ListRecipientProvider for pipeline code

eg:

```
node {
 emailext (
     // attachmentsPattern: '**',
      subject: subject,
      body: details,
      recipientProviders: [
          [$class: 'ListRecipientProvider', recipientList: "${emailList}"]
      ]
   );

}
```

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/recipients/ListRecipientProvider.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/recipients/ListRecipientProvider.java
@@ -16,6 +16,7 @@ import hudson.plugins.emailext.plugins.RecipientProvider;
 import hudson.plugins.emailext.plugins.RecipientProviderDescriptor;
 import jenkins.model.Jenkins;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Makes sense to at least declare the `@Symbolz`to avoid the `$class` definition. 

* I'm not sure this approach is completely safe for AbstractProjects
* Snippet Generator will unlikely work with  such approach (maybe I'm wrong)

qxo:
@oleg-nenashev 
The code work for me in jenkins pipeline.
I have no time to check other use case
You may modify the code for other use case  :)

davidvanlaatum:
I don't understand why this is needed? You can just pass the emails into the emailext command directly eg
```
node {
 emailext (
     // attachmentsPattern: '**',
      subject: subject,
      body: details,
      to: "${emailList}"
   );
}
```

qxo:
@davidvanlaatum  on my test  "to" argument not work:)

davidvanlaatum:
OK log an issue at http://issues.jenkins-ci.org

PR #149, opened 2017-01-23
Added a test result example groovy script. 

Gaoithe:
There are not many examples of email-ext plugin groovy scripts available (google search, google groups, stack overflow, e.t.c.). So here is an example which shows job/build result and expanded test result. Test suite result is shown and individual pass/fail for each test is shown.
 
We found that the default html.jelly worked. But we could not copy it or make changes - the file was found but always encountered a jellyException even with no changes. "JellyException: Could not parse Jelly script : null". No output from jelly in logs - although that may be configurable. Groovy scripting worked but it took a while to get right.

This script I think could be useful to a few people.

### DIFF
--- a/src/main/resources/hudson/plugins/emailext/templates/testresult.groovy
+++ b/src/main/resources/hudson/plugins/emailext/templates/testresult.groovy
@@ -1,0 +1,114 @@
+<html>
+<body>
+<%
+
+    import hudson.model.*
+
+    def build = Thread.currentThread().executable
+    def buildNumber = build.number
+    def buildNumHash = build.getDisplayName()
+
+    def testPassed = "0"
+    def testFailed = "0"
+    def testSkipped = "0"
+    def buildDuration = "0"
+    if(build.testResultAction) {
+        def testResult = build.testResultAction
+        testCount = String.format("%d",(testResult.totalCount))
+        testPassed = String.format("%d",(testResult.result.passCount))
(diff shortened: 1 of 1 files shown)


xinzx:
Dears,

I have a question here, how can I read XML file data to email template groovy script?
I use below code:

```
def demo = new File("/data/jenkins/jobs/carday-at-ie/builds/36/robot-plugin/output.xml")
def sample = new XmlParser().parseText(demo.text)
```

but when I test in Jenkins with "Email Template Testing" function, it always note me as below:
`Exception raised during template rendering: /data/jenkins/jobs/carday-at-ie/builds/36/robot-plugin/output.xml (No such file or directory) java.io.FileNotFoundException:`

actually the file exist on this path.
Is any suggestion? thanks.

dragon788:
@xinzx keep in mind that the path is on the Jenkins server, and the service is running as the jenkins user, so make sure it is readable to that user. The Jenkins service may also be running in a Tomcat "app jail" where it is only aware of its own surroundings and doesn't have access to arbitrary paths on the system. Try using `$JENKINS_HOME` instead of the full path from root, so `$JENKINS_HOME/jobs/etc`.

shailzachaudhry:
Hello,
How can I get my build's maven version in email template groovy script and in the build notification email ?
$POM_VERSION doesn't seem to be working.

davidvanlaatum:
@shailzachaudhry adding unrelated comments will get you ignored

PR #153, opened 2017-03-20
Stop skipping tests during release:prepare

jglick:
Reverts b601cc995f734238d1aac0b8076531039d95cf9a: the parent POM already skips tests during `release:perform`, so this is turning off the valuable safety feature of ensuring that `release:prepare` will not run with outstanding test failures.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -292,12 +292,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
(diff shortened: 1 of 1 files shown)


jglick:
I cannot reproduce test failures locally. Not sure what that is about.

davidvanlaatum:
I did this deliberately as I don't want them to run when I do a release, I have already run them before hand anyway.

PR #155, opened 2017-05-04
[JENKINS-24141] Switch to using RunWithSCM for getCulprits logic

abayer:
[JENKINS-24141](https://issues.jenkins-ci.org/browse/JENKINS-24141)

Downstream of https://github.com/jenkinsci/jenkins/pull/2730 and https://github.com/jenkinsci/workflow-job-plugin/pull/35

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,7 @@
         <workflow-cps.version>2.29</workflow-cps.version>
         <workflow-durable-task-step.version>2.11</workflow-durable-task-step.version>
         <workflow-basic-steps.version>2.4</workflow-basic-steps.version>
+        <workflow-support.version>2.13</workflow-support.version>
         <scm-api.version>2.0.7</scm-api.version>
         <java.level>7</java.level>
         <concurrency>2</concurrency>
@@ -180,6 +181,19 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
             <version>${workflow-api.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-support</artifactId>
+            <version>${workflow-support.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-support</artifactId>
+            <version>${workflow-support.version}</version>
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -32,14 +32,14 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <workflow-step-api.version>2.9</workflow-step-api.version>
         <workflow-aggregator.version>2.5</workflow-aggregator.version>
-        <workflow-job.version>2.12</workflow-job.version>
-        <jenkins.version>2.60</jenkins.version>
+        <workflow-job.version>2.11</workflow-job.version>
+        <jenkins.version>2.0</jenkins.version>
         <workflow-api.version>2.11</workflow-api.version>
         <workflow-cps.version>2.29</workflow-cps.version>
         <workflow-durable-task-step.version>2.11</workflow-durable-task-step.version>
         <workflow-basic-steps.version>2.4</workflow-basic-steps.version>
         <scm-api.version>2.0.7</scm-api.version>
-        <java.level>8</java.level>
+        <java.level>7</java.level>
         <concurrency>2</concurrency>
         <argLine />
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -32,9 +32,8 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <workflow-step-api.version>2.9</workflow-step-api.version>
         <workflow-aggregator.version>2.5</workflow-aggregator.version>
-        <workflow-job.version>2.11-20170504.162002-1</workflow-job.version> <!-- TODO: switch to release once https://github.com/jenkinsci/workflow-job-plugin/pull/35 is merged and released -->
-        <jenkins-core.version>2.58-20170502.192524-8</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-war.version>2.58-20170502.192544-8</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <workflow-job.version>2.12</workflow-job.version>
+        <jenkins.version>2.60</jenkins.version>
         <workflow-api.version>2.11</workflow-api.version>
         <workflow-cps.version>2.29</workflow-cps.version>
         <workflow-durable-task-step.version>2.11</workflow-durable-task-step.version>
--- a/pom.xml
+++ b/pom.xml
@@ -133,7 +133,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>mailer</artifactId>
-            <version>1.5</version>
+            <version>1.13</version>
         </dependency>
         <dependency>
             <groupId>org.jvnet.hudson.plugins</groupId>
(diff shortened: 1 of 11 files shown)


jglick:
Set CI to build with JDK 8.

src/main/java/hudson/plugins/emailext/plugins/recipients/DevelopersRecipientProvider.java
```diff
@@ -39,6 +39,7 @@ public void send(final String format, final Object... args) {
             }
         }
         final Debug debug = new Debug();
+
```
jglick: gratuitous, revert

jglick:
Even assuming the upstream PR does get merged and released in a weekly, there is no expectation this PR would get merged in the near future. Probably you would want to wait until the API makes it into an LTS release, at least.

abayer:
@jglick See https://github.com/jenkinsci/email-ext-plugin/pull/155/files#diff-14528565ae064617b0382642b3e14276 - there was previously special-casing for non-`AbstractBuild`s, which was tested in `CulpritsRecipientProviderTest`, so we just have to make sure the tests keep passing rather than write new ones.

src/main/java/hudson/plugins/emailext/plugins/recipients/DevelopersRecipientProvider.java
abayer: ...dangit. =)

pom.xml
```diff
@@ -30,9 +30,16 @@
     <properties>
         <powermock.version>1.6.4</powermock.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <workflow.version>1.10</workflow.version>
-        <jenkins.version>2.0</jenkins.version>
-        <java.level>7</java.level>
+        <workflow-step-api.version>2.9</workflow-step-api.version>
+        <workflow-aggregator.version>2.5</workflow-aggregator.version>
+        <workflow-job.version>2.12</workflow-job.version>
+        <jenkins.version>2.60</jenkins.version>
```
oleg-nenashev: Needs sign-off from the plugin maintainer

jglick:
> Reflection could help to temporary avoid bumping Jenkins core dependency

Yes, that is definitely an option here, since it is a single method call so the whole idiom would be only a few lines (incl. a comment reminding the maintainer to switch to a direct call when updating the baseline). Would get this in the hands of (2.60+) users sooner, assuming the maintainer is reluctant to use a new baseline, so it might be a reasonable tradeoff.

abayer:
So the reflection thing gets a bit hairier due to there already being a `getChangeSets` on `AbstractBuild` pre-2.60, but it's `@Restricted(DoNotUse.class)`. So basically I'll revert to the previous behavior there.

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
```diff
@@ -51,10 +54,18 @@ public void send(final String format, final Object... args) {
         }
         final Debug debug = new Debug();
         Run<?,?> run = context.getRun();
-        if (run instanceof AbstractBuild) {
-            Set<User> users = ((AbstractBuild<?,?>)run).getCulprits();
-            RecipientProviderUtilities.addUsers(users, context, env, to, cc, bcc, debug);
-        } else {
+        // TODO: core 2.60+, workflow-job 2.12+: Switch to checking if run is RunWithSCM and make catch an else block
```
jglick: …or simply delete the `else` block.

src/main/java/hudson/plugins/emailext/plugins/recipients/RecipientProviderUtilities.java
```diff
@@ -37,6 +37,7 @@
 import hudson.scm.ChangeLogSet;
 import hudson.tasks.MailSender;
 
+import javax.mail.MethodNotSupportedException;
```
jglick: Huh?

pom.xml
```diff
@@ -30,8 +30,15 @@
...
         <powermock.version>1.6.4</powermock.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <workflow.version>1.10</workflow.version>
+        <workflow-step-api.version>2.9</workflow-step-api.version>
+        <workflow-aggregator.version>2.5</workflow-aggregator.version>
+        <workflow-job.version>2.11</workflow-job.version>
         <jenkins.version>2.0</jenkins.version>
+        <workflow-api.version>2.11</workflow-api.version>
+        <workflow-cps.version>2.29</workflow-cps.version>
+        <workflow-durable-task-step.version>2.11</workflow-durable-task-step.version>
```
oleg-nenashev: make sure that all deps are compatible with Jenkins 2.0after the dep downgrade

src/main/java/hudson/plugins/emailext/plugins/recipients/RecipientProviderUtilities.java
```diff
@@ -74,6 +76,7 @@ private RecipientProviderUtilities() {
                     addChangeSetUsers(changeLogSet, users, debug);
                 }
             } else {
+                // TODO: core 2.60+, workflow-job 2.12+: Decide whether to remove this logic since it won't be needed for Pipelines any more.
```
jglick: Right, in 2.60+ we can just drop the `else` clause.

src/test/java/hudson/plugins/emailext/plugins/recipients/FirstFailingBuildSuspectsRecipientProviderTest.java
```diff
@@ -117,9 +121,9 @@ public void testAddRecipients() throws Exception {
          * Committers {U,V}.
          * Failed.
          */
-        final FreeStyleBuild build5 = PowerMockito.mock(FreeStyleBuild.class);
-        PowerMockito.when(build5.getPreviousCompletedBuild()).thenReturn(build4);
-        PowerMockito.when(build5.getResult()).thenReturn(Result.FAILURE);
+        final FreeStyleBuild build5 = PowerMockito.spy(new FreeStyleBuild(p));
+        PowerMockito.doReturn(Result.FAILURE).when(build5).getResult();
+        PowerMockito.doReturn(build4).when(build5).getPreviousCompletedBuild();
```
jglick: This is one reason I loathe mocking tests—they have to be rewritten even if you make _compatible_ changes.

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
```diff
@@ -51,10 +54,18 @@ public void send(final String format, final Object... args) {
...
+            Method getCulprits = run.getClass().getMethod("getCulprits");
+            if (Set.class.isAssignableFrom(getCulprits.getReturnType())) {
+                @SuppressWarnings("unchecked")
+                Set<User> users = (Set<User>) getCulprits.invoke(run);
+                if (Iterables.all(users, Predicates.instanceOf(User.class))) {
+                    RecipientProviderUtilities.addUsers(users, context, env, to, cc, bcc, debug);
+                }
+            }
+        } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
+            debug.send("Exception getting culprits for %s: %s", run, e);
```
oleg-nenashev: 🐜 this is a valid behavior for 2.60-, no? Maybe it needs more graceful message, but up to you

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
abayer: Yeah, I guess we can probably just rule out any other weirdo job types out there.

src/main/java/hudson/plugins/emailext/plugins/recipients/RecipientProviderUtilities.java
abayer: Whoops..

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
abayer: Actually, I like what I've got here. It's a nice better-safe-than-sorry fallback.

src/test/java/hudson/plugins/emailext/plugins/recipients/FirstFailingBuildSuspectsRecipientProviderTest.java
abayer: Ahyup.

pom.xml
abayer: They are - none of the core Pipeline plugins have dependencies on core >=2.0 until `workflow-job` 2.12.

src/main/java/hudson/plugins/emailext/plugins/recipients/CulpritsRecipientProvider.java
abayer: Good point.

jglick:
Test failures which look real.

PR #158, opened 2017-09-11
Correct use of build.changeSets

ssbarnea:
Fixes wrong variable name in template
which was failing rendering of template.

JENKINS-38968

### DIFF
--- a/src/main/resources/hudson/plugins/emailext/templates/groovy-html.template
+++ b/src/main/resources/hudson/plugins/emailext/templates/groovy-html.template
@@ -17,8 +17,8 @@ TD.console { font-family:Courier New; }
 <BODY>
 
 <TABLE>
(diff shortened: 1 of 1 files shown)


ssbarnea:
@davidvanlaatum Thanks for the very quick response. I am doing the testing now in order to see if it works, obviously that we should not merge it until I have the confirmation that it works.

Sadly it seems that CI is broken on master, https://ci.jenkins.io/job/Plugins/job/email-ext-plugin/

ssbarnea:
I am unable to confirm or infirm the fix because the current result with the new version is:
* there is no exception anymore in code
* the changeset is table is empty, but this may be because I tested a build on master branch, so not necessarily a bug
* email came in HTML as text, so is not properly encoded 

No idea what to do abut making the email as HTML instead of plain text, but until I sort this issue I cannot give the ok.

ssbarnea:
I found https://support.cloudbees.com/hc/en-us/articles/226237768-How-to-embed-html-report-in-email-body-using-Email-ext-plugin- but it doe not help in my case as do have the HTML as a string, not as a file on disk..

ssbarnea:
Good news! The change is working fine. There is only one bug but is unrelated to this change: the build status appears as null on the jobs:

See http://s3.sbarnea.com/ss/170911-1556.png

Again, this is unrelated to this change and the change is ready to be merged. It does fix

PR #159, opened 2017-09-11
JENKINS-33574 fix unicode attachments

ssbarnea:


davidvanlaatum:
Looks like it only fails on windows which is a pain since I run osx and linux. Passes locally for me but can see it failed on the jenkins ci server when run on windows

ssbarnea:
This change is not fixing the bug, only making the test output more verbose so we can debug it. Still, I was able to find more information about this annoying bug.

It does fail even on my MacOS High Sierra but if I add "LANG=en_US.UTF-8 mvn package" and run it i does pass. Sounds like an issue with java or maven and we need to find a way to bypass it.

Can you try to run it with these vars?
```
export MAVEN_OPTS='-Dorg.slf4j.simpleLogger.log.org.apache.maven.cl‌​i.transfer.Slf4jMave‌​nTransferListener=wa‌​rn'
LANG= LC_CTYPE= mvn package
```
Maybe this would allow us to narrow down and find a way to patch `pom.xml` or `Jenkinsfile` to make it pass. As last resort we can annotate the test and make it skip run on Windows platform.

davidvanlaatum:
Yes that makes it fail for me

ssbarnea:
So `LANG=` is what it breaks it. I just pushed some changes (found on stackoverflow) that may bypass this issue. Lets wait for their results. I know that Windows has Unicode support but is very different than Unix and I suspect that we get the failure due to the console encoding and not due to the Java and/or filesystem limitations. 

I worked as internationalization engineer for 4 years so I had to do a lot of work around Unicode in the past, but that was like 6+ years ago. So glad I do not have to deal with Windows anymore.

ssbarnea:
@davidvanlaatum Time to merge it! It did fix the build. And after this you should be able to make a new bug fix release.

davidvanlaatum:
done

PR #162, opened 2017-11-30
Styling changes + change in test results table + code alignment

rdagum:
Sample of success:
image

Sample of failure:
image


davidvanlaatum:
Would also have been nice to include a before image just so I didn't have to go hunting to figure out what it looked like before ;)

davidvanlaatum:
Also please include a third color for unstable

rdagum:
I updated to use same styling in artifacts and maven artifacts. However, I couldn't test the maven artifacts output because I don't have a project that uses that.
I added the test age as well.

davidvanlaatum:
Just for your own info http://github.com/jglick/simple-maven-project-with-tests

PR #166, opened 2018-02-15
Add additional accounts

Kosterio:
Added ability of additional smtp accounts and java mail parameters


src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
```diff
@@ -577,23 +583,36 @@ public boolean configure(StaplerRequest req, JSONObject formData)
...
+                addAccounts.add(new MailAccount((JSONObject)addacc));
+            }
+        }
+
         // specify authentication information
         if (req.hasParameter("ext_mailer_use_smtp_auth")) {
-            smtpAuthUsername = nullify(req.getParameter("ext_mailer_smtp_username"));
-            smtpAuthPassword = Secret.fromString(nullify(req.getParameter("ext_mailer_smtp_password")));
+            mailAccount.setSmtp_username(nullify(req.getParameter("ext_mailer_smtp_username")));
+            mailAccount.setSmtp_password(nullify(req.getParameter("ext_mailer_smtp_password")));
```
davidvanlaatum: Why the loss of Secret.fromString

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
```diff
@@ -679,13 +698,13 @@ void upgradeFromMailer() {
         // get the data from Mailer and then set override to true
         this.defaultSuffix = Mailer.descriptor().getDefaultSuffix();
         this.defaultReplyTo = Mailer.descriptor().getReplyToAddress();
-        this.useSsl = Mailer.descriptor().getUseSsl();
+        mailAccount.setUse_ssl(Mailer.descriptor().getUseSsl());
         if (StringUtils.isNotBlank(Mailer.descriptor().getSmtpAuthUserName())) {
-            this.smtpAuthPassword = Secret.fromString(Mailer.descriptor().getSmtpAuthPassword());
-            this.smtpAuthUsername = Mailer.descriptor().getSmtpAuthUserName();
+            mailAccount.setSmtp_username(Mailer.descriptor().getSmtpAuthUserName());
+            mailAccount.setSmtp_password(Mailer.descriptor().getSmtpAuthPassword());
```
davidvanlaatum: Again

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
Kosterio: because it makes inside of class

davidvanlaatum:
Please contact me on my email which is in the pom.xml I would like to talk about something off the record.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
davidvanlaatum: That is by design and is required to store the password encrypted

davidvanlaatum: oh sorry missed it in the subclass

src/main/java/hudson/plugins/emailext/MailAccount.java
```diff
@@ -0,0 +1,99 @@
...
+        public String getDisplayName(){
+            return "";
+        }
+    }
+
+    public String getAddress(){
+        return address;
+    }
+
+    public String getSmtp_host(){
```
davidvanlaatum: should really be getSmtpHost ie camelCase not a mix of camelCase and snake_case

davidvanlaatum:
also noticed there is no Use SMTP Authentication checkbox for additional accounts.

davidvanlaatum:
can't believe I missed this before but also need to add some unit tests

Kosterio:
sorry for rebase

davidvanlaatum:
I'd like to see a test for create session if possible to assert that it used the right account, probably easiest to call it and then inspect the javamail properties on the returned session to see which server it used

davidvanlaatum:
how are we going with this? I am waiting on this for the next release

Kosterio:
I don't know how to create mock session and test her. I had not free time to figure it out.

davidvanlaatum:
You probably don't need to create a mock one I had a look and you should be able to call the function then get the properties off it and check it used the correct one

espoelstra:
@davidvanlaatum is it possible this broke the `setSmtpAuth` by overriding the MailAccount that was previously inherited from Mailer with the new one that is included in this PR? Suddenly our Groovy configuration script that was working with a previous version started throwing errors. This is the interface we were configuring through, https://github.com/jenkinsci/email-ext-plugin/blob/master/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java#L346

davidvanlaatum:
https://issues.jenkins-ci.org/browse/JENKINS-52949

PR #167, opened 2018-04-24
allow filtering email domains we send emails to

ArturHarasimiuk:
This implements domain filter where we specify all alowed domains. prior
to sending email we check if destination address is on white list. if
not then email address is removed from the list.


src/main/webapp/help/globalConfig/allowedDomains.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    Allows to restrict list of destination domains where we want send emails to. You can specify more than one domain by using comma separator.<br/>
+    <code>Example: @domain1.com,domain2.com<code>
```
davidvanlaatum: Please add a comment around the difference between including the @ or not

PR #170, opened 2018-07-26
JENKINS-52748: Same recipient in CC or BCC removes it from TO

reinholdfuereder:
This PR addresses https://issues.jenkins-ci.org/browse/JENKINS-52748

### DIFF
--- a/src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
+++ b/src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
@@ -29,15 +29,20 @@ public class EmailRecipientUtils {
     
     public static Set<InternetAddress> convertRecipientString(String recipientList, EnvVars envVars)
             throws AddressException, UnsupportedEncodingException {
-        return convertRecipientString(recipientList, envVars, TO);
+        return convertRecipientString(recipientList, envVars, TO, null);
     }
-    
+
     public static Set<InternetAddress> convertRecipientString(String recipientList, EnvVars envVars, int type)
+            throws AddressException, UnsupportedEncodingException {
+        return convertRecipientString(recipientList, envVars, type, null);
+    }
+
+    private static Set<InternetAddress> convertRecipientString(String recipientList, EnvVars envVars, int type, TaskListener listener)
         throws AddressException, UnsupportedEncodingException {
         final Set<InternetAddress> internetAddresses = new LinkedHashSet<>();
         if (!StringUtils.isBlank(recipientList)) {
             final String expandedRecipientList = fixupDelimiters(envVars.expand(recipientList));
-            RecipientListStringAnalyser recipientsAnalyser = new RecipientListStringAnalyser(expandedRecipientList);
+            RecipientListStringAnalyser recipientsAnalyser = RecipientListStringAnalyser.newInstance(listener, expandedRecipientList);
             InternetAddress[] all = InternetAddress.parse(expandedRecipientList.replace("bcc:", "").replace("cc:", ""));
             final Set<InternetAddress> to = new LinkedHashSet<>();
             final Set<InternetAddress> cc = new LinkedHashSet<>();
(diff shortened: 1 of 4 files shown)


reinholdfuereder:
Hm, check has failed with IMHO unrelated tests running into timeout; blue ocean writes "existing failures" (instead of "new" or so)!?

src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
```diff
@@ -38,28 +38,26 @@
...
+            String[] recipients = StringUtils.stripAll(expandedRecipientList.split(","));
+            if (recipients.length != all.length) {
+                throw new IllegalArgumentException("Unsupported format in expanded recipient list string: " + expandedRecipientList);
+            }
             final Set<InternetAddress> to = new LinkedHashSet<>();
             final Set<InternetAddress> cc = new LinkedHashSet<>();
             final Set<InternetAddress> bcc = new LinkedHashSet<>();
             final String defaultSuffix = ExtendedEmailPublisher.descriptor().getDefaultSuffix();
 
-            for(InternetAddress address : all) {
```
davidvanlaatum: I think what this code was trying to support (badly) is the weird format some stuff returns email address in ie
`"Mouse, Mickey" <mickey@disney.com>`
I just added a test case for this and it breaks with your changes

davidvanlaatum: But thanks for taking the time to try and fix

reinholdfuereder: Oh! That explains the seemingly overly complex/confusing existing (and with respect to JENKINS-52748 buggy) code: thanks for the hint and the new test case. This does not make it really easier ;-)
I might have another look and try at this tomorrow; unless you already have an idea that you want to implement...

reinholdfuereder:
Still failing test is IMHO unrelated to my changes and due to https://issues.jenkins-ci.org/browse/JENKINS-32510

reinholdfuereder:
@davidvanlaatum Would you mind having now a closer look at this PR again?

- Your recently added new test succeeds.
- My newly added test to demonstrate the fix for JENKINS-52748 succeeds.
- Unfortunately due to JENKINS-32510 same tests that are unrelated to my changes randomly/sporadically fail due to running into timeouts.
- I tried to make reviewing easier by making many commits with small changes, as I e.g. also tried to improve the existing test cases of `EmailRecipientUtilsTest`; so maybe instead of just looking at the final result it may be good to look at individual commits (first).

src/main/java/hudson/plugins/emailext/RecipientListStringAnalyser.java
```diff
@@ -0,0 +1,101 @@
+package hudson.plugins.emailext;
+
+import javax.mail.internet.InternetAddress;
+import java.text.MessageFormat;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+class RecipientListStringAnalyser {
+
+    private static final Logger LOG = Logger.getLogger(RecipientListStringAnalyser.class.getName());
```
davidvanlaatum: I know its not so nice but there is a debug logger thingy used else ware that enables switching on debugging in settings so debug logs likes these end up in the build output. If its not too hard would prefer to use that

reinholdfuereder: (Sorry for the delay)
After some looking/searching I guess you refer to the "Configure System" setting "Enable Debug Mode" for "Extended E-mail plugin"? => I need to use `ExtendedEmailPublisherContext` and `ExtendedEmailPublisherDescriptor`?

reinholdfuereder: Well, after much more looking I dare to say that there is one code dependency that I think does not allow logging via that approach:
* Dependencies:
```
RecipientListStringAnalyser
  static hudson.plugins.emailext.EmailRecipientUtils#convertRecipientString(java.lang.String, hudson.EnvVars, int)
    static hudson.plugins.emailext.EmailRecipientUtils#convertRecipientString(java.lang.String, hudson.EnvVars)
    hudson.plugins.emailext.EmailRecipientUtils#validateFormRecipientList
      hudson.plugins.emailext.ExtendedEmailPublisherDescriptor#doRecipientListRecipientsCheck
    static hudson.plugins.emailext.EmailRecipientUtils#addAddressesFromRecipientList
```
  * My new `RecipientListStringAnalyser` class is only used by `EmailRecipientUtils` class
  * But this in turn is used mostly statically, but not always; and also does not always have an `ExtendedEmailPublisherContext` instance or its task listener (which in turn holds the logger): namely the problem is in the `hudson.plugins.emailext.EmailRecipientUtils#validateFormRecipientList` dependency

The resulting code will not be very nice, because the logging must be conditionally on the availability of the logger and so no logging for `hudson.plugins.emailext.EmailRecipientUtils#validateFormRecipientList`/`hudson.plugins.emailext.ExtendedEmailPublisherDescriptor#doRecipientListRecipientsCheck` is possible, I think.

Since I don't know what this is used for I am not sure if I should try that (non-perfect/-elegant approach via the conditionally logging)?

reinholdfuereder: Oh, and I would need to break compatibility of `hudson.plugins.emailext.EmailRecipientUtils`, I think; but maybe that is OK anyhow?

davidvanlaatum:
yea there are some big bits I would like to rewrite the gathering of email addresses could be improved a lot as well as adding other transport types at the same time (eg slack) but my health is just not up to it atm which is why I have put out a request for a co maintainer.

reinholdfuereder:
Recent failing check (https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Femail-ext-plugin/detail/PR-170/8/pipeline/) is IMHO caused due to extreme build slowness and thus running into build timeout and thus build being killed?

@davidvanlaatum Can you please re-trigger the build? Or do I have to make a dummy commit + push?

davidvanlaatum:
I can't trigger but lets see if its set for comments
jenkins please retry build
Also reduced the concurrency

reinholdfuereder:
@davidvanlaatum Thanks!

PR #175, opened 2018-12-21
[JENKINS-55233] Make sure to use latest pom and recommend lib versions

alecharp:
Using pom `3.32` and recommended version for powermock solve the problem.

@jenkinsci/java11-support @davidvanlaatum

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(failFast: false)
+buildPlugin(failFast: false, jdkVersions: [8, 11])
--- a/pom.xml
+++ b/pom.xml
@@ -4,8 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <!-- TODO: use the release asap -->
-        <version>3.33-SNAPSHOT</version>
+        <version>3.32</version>
         <relativePath />
     </parent>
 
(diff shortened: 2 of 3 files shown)


slide:
Any idea on why the tests are failing?

alecharp:
Not a clue. I don't have the same behaviour locally. Don't merge, I'll work
on it.

alecharp:
I have the same issue locally. I'm working on solving that.

alecharp:
So, I discovered that with the new version of PowerMock, it seems that the real method `get` is called rather than the "interceptor" which is set in `MockUtilities#addRequestor` on line 131-137. 

First I thought to blame the `RC.4` of the libraries (so I did jenkinsci/plugin-pom/pull/148) but it is not solving  the problem.

alecharp:
From `Mockito.anyMap()` javadoc, it is not matching `null` anymore. 

> Since Mockito 2.1.0, only allow non-null <code>Map</code>. 
As this is a nullable reference, the suggested API to <strong>match</strong> <code>null</code> wrapper would be {@link #isNull()}. We felt this change would make tests harness much safer that it was with Mockito 1.x.

So replacing the `anyMap` by `any` solves the problem. I have the same `null` issue when using `any(Map.class)`, I  cannot be really specific about what the third argument should be.

Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(failFast: false)
+buildPlugin(failFast: false, jdkVersions: [8, 11])
```
batmat: :+1:

PR #176, opened 2019-02-11
Fix build configuration for Java 11 validation

alecharp:
Currently, the build of Master is failing because of the Java 11 branches configuration.

This is a follow-up from jenkinsci/email-ext-plugin#175 were I introduce the build
on Java 11 but misconfigured it. Sorry for that.

@jenkinsci/java11-support

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,9 @@
-buildPlugin(failFast: false, jdkVersions: [8, 11])
+buildPlugin(
+    failFast: false,
+    configurations: [
+        [ platform: 'linux', jdk: "8", jenkins: null ],
+        [ platform: 'windows', jdk: "8", jenkins: null ],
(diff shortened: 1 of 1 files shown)


slide:
Should this succeed?

batmat:
@slide yes, but it won't as is, because given @alecharp has no merge access here, anyway his PR is getting ignored.

alecharp:
@slide I wish, but as I don't have write permission on this repository, the `Jenkinsfile` I submit is not used to build the changeset of the PR. So it is still using the `Jenkinsfile` from `master` branch.

batmat:
I think this should either be filed by Alex, or Oleg given he had admin permission, hence it would use the Jenkinsfile from the PR in such case.

alecharp:
I agree, this way the change would still be in a PR if it requires adjustments.

slide:
Superseded by #177

PR #178, opened 2019-03-15
Email -> E-mail in config page

VirtualTim:
Trivial change, but it's referred to as E-mail in the rest of the plugin.


VirtualTim:
I think I've changed it in all the places I need to.

PR #179, opened 2019-03-20
Use recommended build configuration + fix FindBugs issues

batmat:
See https://github.com/jenkins-infra/pipeline-library/pull/86

Should supersede #177

PR #180, opened 2019-03-21
Java11 readiness

oleg-nenashev:
Resubmitting #179 with repo WRITE superpowers

CC @jenkinsci/java11-support


batmat:
So my Jenkinsfile is not used, but the one on `master`, so I believe the error is expected...

PR #188, opened 2019-07-07
Reduce deprecated API usage

darxriggs:


slide:
Close and reopen to get new build

src/test/java/hudson/plugins/emailext/EmailExtStepTest.java
```diff
@@ -127,7 +127,7 @@ public void saveOutput() throws Exception {
                 "node {\n" +
                         "  emailext(to: 'mickeymouse@disney.com', subject: 'Boo', saveOutput: true)\n" +
                         "  archiveArtifacts '*.*'\n" +
-                        "}", false));
```
slide: Is there a specific reason to change this from false to true?

darxriggs: Mainly for consistency reasons. In all other places in this test, the `sandbox` parameter wasn't used before and now is set to `true`. Therefore I also changed this one. This therefore also makes sure that the plugin runs fine without any script approval inside the sandbox.

PR #194, opened 2019-07-30
Added escapeHtml flag to FAILED_TESTS

lukaszkostrzewa:
This PR adds flag for escaping characters in errors details and stack traces using HTML entities in `FAILED_TESTS` plugin macro token.

Examples:
```
expected:<ABORTED> but was:<COMPLETED>
at org.nexusformat.NexusFile.<clinit>(NexusFile.java:99)
```
will be replaced to:
```
expected:&lt;ABORTED&gt; but was:&lt;COMPLETED&gt;
at org.nexusformat.NexusFile.&lt;clinit&gt;(NexusFile.java:99)
```

It will be useful when content type is set to text/html.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -13,9 +13,11 @@ import org.jenkinsci.plugins.tokenmacro.MacroEvaluationException;
 
 import java.io.IOException;
 
+import static org.apache.commons.lang.StringEscapeUtils.escapeHtml;
+
 /**
  * An EmailContent for failing tests. Only shows tests that have failed.
- * 
+ *
  * @author markltbaker
  */
 @Extension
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
```diff
@@ -119,11 +124,13 @@ private int outputTest(StringBuilder buffer, TestResult failedTest,
...
-            local.append("\nError Message:\n").append(failedTest.getErrorDetails()).append('\n');
+            String errorDetails = escapeHtml ? escapeHtml(failedTest.getErrorDetails()) : failedTest.getErrorDetails();
+            local.append("\nError Message:\n").append(errorDetails).append('\n');
         }
-        
+
         if (showStack) {
-            local.append("\nStack Trace:\n").append(failedTest.getErrorStackTrace()).append('\n');
+            String stackTrace = escapeHtml ? escapeHtml(failedTest.getErrorStackTrace()) : failedTest.getErrorStackTrace();
+            local.append("\nStack Trace:\n").append(stackTrace).append('\n');
```
slide: When using escapeHtml, would you want to use `<br/>` instead of `\n` in the strings?

lukaszkostrzewa: Good point, I've changed it.

PR #195, opened 2019-10-17
Improvements for pom.xml

darxriggs:


slide:
Has some test failures

darxriggs:
They have been fixed now. Along the way surefire in combination with JaCoCo was fixed too.

PR #196, opened 2020-01-29
[JENKINS-60911] Make files (template, script, etc...) correctly loaded through config-file-provider-plugin when defined at folder level

gcernier-amadeus:
Fix of [JENKINS-60911](https://issues.jenkins-ci.org/browse/JENKINS-60911) issue.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
@@ -24,7 +24,6 @@
 package hudson.plugins.emailext.plugins.content;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.Plugin;
 import hudson.model.AbstractBuild;
@@ -40,6 +39,7 @@ import jenkins.security.NotReallyRoleSensitiveCallable;
 import org.apache.commons.io.FilenameUtils;
 import org.jenkinsci.lib.configprovider.ConfigProvider;
 import org.jenkinsci.lib.configprovider.model.Config;
+import org.jenkinsci.plugins.configfiles.ConfigFiles;
 import org.jenkinsci.plugins.scriptsecurity.scripts.Language;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.tokenmacro.DataBoundTokenMacro;
@@ -54,6 +54,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -95,14 +96,14 @@ public abstract class AbstractEvalContent extends DataBoundTokenMacro {
         return false;
     }
     
-    protected InputStream getFileInputStream(FilePath workspace, String fileName, String extension)
+    protected InputStream getFileInputStream(Run<?, ?> run, FilePath workspace, String fileName, String extension)
             throws FileNotFoundException, IOException, InterruptedException {
         
         InputStream inputStream = null;
         if(fileName.startsWith("managed:")) {
             String managedFileName = fileName.substring(8);
             try {
-                inputStream = getManagedFile(managedFileName);
+                inputStream = getManagedFile(run, managedFileName);
             } catch(NoClassDefFoundError e) {
                 inputStream = null;
             }
@@ -159,15 +160,13 @@ public abstract class AbstractEvalContent extends DataBoundTokenMacro {
         return parent.act(new IsChildFileCallable(potentialChild));
     }
 
-    private InputStream getManagedFile(String fileName) throws UnsupportedEncodingException {
+    private InputStream getManagedFile(Run<?, ?> run, String fileName) throws UnsupportedEncodingException {
         InputStream stream = null;
         Plugin plugin = Jenkins.get().getPlugin("config-file-provider");
         if (plugin != null) {
             Config config = null;
-            ExtensionList<ConfigProvider> providers = ConfigProvider.all();
-            ConfigProvider provider = providers.get(getProviderClass());
-            assert provider != null;
-            for (Config c : provider.getAllConfigs()) {
+            List<Config> configs = ConfigFiles.getConfigsInContext(run.getParent().getParent(), getProviderClass());
+            for (Config c : configs) {
                 if (c.name.equalsIgnoreCase(fileName)) {
                     config = c;
                     break;
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
 
     <properties>
         <java.level>8</java.level>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.138.4</jenkins.version>
         <hpi.compatibleSinceVersion>2.57.2</hpi.compatibleSinceVersion>
         <workflow-step-api.version>2.19</workflow-step-api.version>
         <workflow-aggregator.version>2.5</workflow-aggregator.version>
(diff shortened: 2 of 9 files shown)


src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
```diff
@@ -159,15 +160,14 @@ public static boolean isChildOf(final FilePath potentialChild, final FilePath pa
...
         InputStream stream = null;
         Plugin plugin = Jenkins.get().getPlugin("config-file-provider");
         if (plugin != null) {
             Config config = null;
-            ExtensionList<ConfigProvider> providers = ConfigProvider.all();
-            ConfigProvider provider = providers.get(getProviderClass());
-            assert provider != null;
-            for (Config c : provider.getAllConfigs()) {
+            assert getProviderClass() != null;
+            List<Config> configs = ConfigFiles.getConfigsInContext(run.getParent().getParent(), getProviderClass());
```
slide: Does this work for ALL cases? Why .getParent().getParent()? What if there is folder nesting, does it work in that case?

src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
```diff
@@ -159,15 +160,14 @@ public static boolean isChildOf(final FilePath potentialChild, final FilePath pa
...
+    private InputStream getManagedFile(Run<?, ?> run, String fileName) throws UnsupportedEncodingException {
         InputStream stream = null;
         Plugin plugin = Jenkins.get().getPlugin("config-file-provider");
         if (plugin != null) {
             Config config = null;
-            ExtensionList<ConfigProvider> providers = ConfigProvider.all();
-            ConfigProvider provider = providers.get(getProviderClass());
-            assert provider != null;
-            for (Config c : provider.getAllConfigs()) {
+            assert getProviderClass() != null;
```
slide: Since this is called below, it would be good to cache this so its not called twice when it doesn't need to be.

src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
gcernier-amadeus: Yes `run.getParent().getParent()` is just the way to obtain the `run`'s job and then the job's "container" (called `ItemGroup`), whether it's a nested subfolder or the top level.

By the way, that `provider.getAllConfigs()` usage is deprecated since long time, and [`ConfigFiles.getConfigsInContext(...)` usage](https://github.com/jenkinsci/config-file-provider-plugin/blob/master/src/main/java/org/jenkinsci/plugins/configfiles/ConfigFiles.java#L48) suggested as replacement [since 3 years](https://github.com/jenkinsci/config-file-provider-plugin/commit/bba752a45fa7c35824dcf516df1b270f4c658a5f#diff-c8a9d185f81b3702118c7cccc1c5c98cR133) ; the context here being the folder (the previously mentioned `ItemGroup`) from where the execution occurs.

Note that, therefore, [all existing cases already unit-tested passed](https://ci.jenkins.io/job/Plugins/job/email-ext-plugin/job/PR-196/lastCompletedBuild/testReport/hudson.plugins.emailext.plugins.content/).

src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
gcernier-amadeus: In fact, I realize that this `assert` is not very relevant, since it was not performed before: it was the `provider` instance for which there was a null-check ; not its class (and not anymore `provider` instance is involved with this new usage).
Are you fine if I remove this useless assertion, then?

gcernier-amadeus: FYI @slide, I removed this assertion and pushed the change.

slide:
Running this locally, I am seeing this issue in `testTemplateFromConfigWithinFolderShouldOverrideTemplateFromGlobalConfig`, I am not sure why it fails locally, but works on ci.j.io

```
org.junit.ComparisonFailure: 
Expected :HELLO WORLD from global config!
Actual   :Groovy Template file [managed:email-ext-template-from-global-config.template] was not found in $JENKINS_HOME/email-templates.
<Click to see difference>


	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at hudson.plugins.emailext.plugins.content.ScriptContentTest.testTemplateFromConfigWithinFolderShouldOverrideTemplateFromGlobalConfig(ScriptContentTest.java:161)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:600)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

gcernier-amadeus:
It's indeed really strange to fail only on your local environment and not on build/CI machine (neither on my local environment).
But the stacktrace you pasted shows an assertion failure at [line 161 of ScriptContentTest](https://github.com/jenkinsci/email-ext-plugin/pull/196/files#diff-8dff3614387157a4b280c839f89db64cR161), which does not correspond to any unit-test assertion (not even a line of code).

Any local modification?

gcernier-amadeus:
Indeed @slide, it's really strange to fail only on your local environment and not on build/CI machine (neither on my local environment).
But the stacktrace you pasted shows an assertion failure at [line 161 of ScriptContentTest](https://github.com/jenkinsci/email-ext-plugin/pull/196/files#diff-8dff3614387157a4b280c839f89db64cR161), which does not correspond to any unit-test assertion (not even a line of code).

Any local modification?

slide:
I added a test case, I'll remove it to make sure it's not impacting anything.

slide:
Ok, the issue is that some of the file paths are too long for Windows so that config.xml file is not created on the file system. The Windows on ci.j.io probably have the long path option enabled already. I will enable it and recheck just to make sure. One thing I would like to see in this just for full coverage is a test with a nested folder structure. You can specify a parent-folder with the template and then a config.xml at the test-folder level with no properties. This would make the testing more complete in my mind.

slide:
Once I enabled long paths on Windows and in Git, the tests all passed. Can you add one test that has nested folders to fill out the testing paths.

Something like:

```
parent-folder
    test-folder
        job
```

You can put the template at the parent-job folder and then also do a test where there is a template in test-folder as well.

Let me know if that makes sense or not.

gcernier-amadeus:
Ok, I will add such a test and let you know.

gcernier-amadeus:
Here it is @slide.
In fact, I merged the test case you described with the existing ones I created to a single test case covering the same scenarii.
Note that, this actually re-tests the `config-file-provider-plugin` resolution behaviour when using `Folder` hierarchy ([`cloudbees-folder-plugin`](https://github.com/jenkinsci/cloudbees-folder-plugin)) which is already tested through [their `org.jenkinsci.plugins.configfiles.folder. FolderConfigFileActionTest#correctFileMustBeSelectedInHierarchy` testcase](https://github.com/jenkinsci/config-file-provider-plugin/blob/master/src/test/java/org/jenkinsci/plugins/configfiles/folder/FolderConfigFileActionTest.java#L144).

But it's indeed safer to do it here as well :)

gcernier-amadeus:
Thanks @slide for the approval and merge.
Any idea of when a new version, therefore integrating this fix, will be released?

slide:
I'll get to it early next week

PR #197, opened 2020-02-05
Consider ALL upstream causes in upstream recipient provider

cpt1gl0:
The current implementation of upstream recipient provider will not consider all upstream causes if a run has more than one upstream cause. In order to consider all causes I changed the implementation to find all causes recursively.


slide:
Not building right now, please look at the build failure.

cpt1gl0:
Finally. All checks passed.

slide:
Would it be possible to add a test that covers the case you are talking about?

basil:
Closing in favor of #254.

PR #202, opened 2020-03-07
JENKINS-61379 - Run presend script before checking for empty recipient list

SKroell:
When sending emails using the email-ext plugin, I'm manually populating the recipient list using the pre-send script feature. This means that the recipient list is empty at first and gets populated during the pre-send step.

Unfortunately an empty recipient list stops the entire process before the pre-send step is even run, with the message: "An attempt to send an e-mail to empty list of recipients, ignored".

It would make sense to let the pre-send script run before checking if the recipient list is empty.


SKroell:
Turns out there are specific reasons for the current ordering. 
Current solution is to simply have a dummy entry, that is removed in the presend script.
https://issues.jenkins-ci.org/browse/JENKINS-61379

Closing

slide:
I'm going to reopen this and review it some more.

PR #203, opened 2020-03-07
JENKINS-61380

SKroell:
https://issues.jenkins-ci.org/browse/JENKINS-61380

When using the rebuild plugin and configuring email notifications to use the RequestorRecipientProvider, the email-ext plugin will email the original upstream build requestor and not the user who requested the rebuild.

Either RequestorRecipientProvider should take rebuilds into account somehow
or a BuildUserRecipientProvider (Maybe named differently) should be provided to notify the requestor of the current build without taking upstream causes into account.


slide:
I'm not sure "Build User" is descriptive enough, but I am not sure what else to call it.

SKroell:
I agree that it's not very descriptive. Maybe something like
"Requestor (Ignore Upstream)"? or rename the current "Requestor" to "Requestor (Upstream)"?

PR #207, opened 2020-05-13
Add additional parameter to `FAILED_TESTS` token to enable yaml output generation

DivyeshChandra:
`FAILED_TESTS` token is often used to generate emails of build failures. 
The output is currently a plain text file. It is very hard to iterate through this text file to read the failures case by case. Hence, a format easily readable and iterable would be beneficial. 
YAML is a human readable as well as machine readable format.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -212,7 +212,6 @@ public class FailedTestsContent extends DataBoundTokenMacro {
         public String name;
         public String errorMessage;
         public String stackTrace;
-        public FailedTest() {}
         public FailedTest(String name, String errorMessage, String stackTrace){
             this.errorMessage = errorMessage;
             this.name = name;
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -50,7 +50,7 @@ public class FailedTestsContent extends DataBoundTokenMacro {
     public boolean escapeHtml = false;
 
     @Parameter
-    public String outputYaml = "";
+    public String outputFormat = "";
 
     @Parameter
     public String testNamePattern = "";
@@ -73,7 +73,7 @@ public class FailedTestsContent extends DataBoundTokenMacro {
         AbstractTestResultAction<?> testResult = run.getAction(AbstractTestResultAction.class);
         SummarizedTestResult result = prepareSummarizedTestResult(testResult);
 
-        switch (outputYaml) {
+        switch (outputFormat) {
             case "yaml":
                 try {
                     return result.toYamlString();
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -219,6 +219,11 @@ public class FailedTestsContent extends DataBoundTokenMacro {
             this.name = name;
             this.stackTrace = stackTrace;
         }
+
+        @Override
+        public String toString() {
+            return String.format("Name:%s, Error message: %s, Stack trace:%s", this.name, this.errorMessage, this.stackTrace);
+        }
     }
 
     @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
@@ -245,6 +250,16 @@ public class FailedTestsContent extends DataBoundTokenMacro {
 
         public void addTest(FailedTest t) {
             this.tests.add(t);
+        }
+
+        @Override
+        public String toString() {
+            StringBuilder b = new StringBuilder();
+            for(FailedTest t: this.tests) {
+                b.append(t.toString()).append("\n");
+            }
+            return String.format("Summary:%s, Tests: %s, Other failed tests:%s, Output truncated: %s",
+                    this.summary, b.toString(), this.otherFailedTests, this.truncatedOutput);
         }
     }
 }
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -112,24 +112,25 @@ public class FailedTestsContent extends DataBoundTokenMacro {
         else {
             result.summary = String.format("%d tests failed.", failCount);
             setMaxLength();
-            if(maxTests == 0) return result;
-            int printSize = 0;
-            for (TestResult failedTest : testResult.getFailedTests()) {
-                if (regressionFilter(failedTest)) {
-                    String stackTrace = showStack ? (escapeHtml ? escapeHtml(failedTest.getErrorStackTrace()) : failedTest.getErrorStackTrace()) : null;
-                    String errorDetails = showMessage ? (escapeHtml ? escapeHtml(failedTest.getErrorDetails()) : failedTest.getErrorDetails()) : null;
-                    String name = String.format( "%s.%s", (failedTest instanceof CaseResult) ? ((CaseResult)failedTest).getClassName() : failedTest.getFullName(),
-                            failedTest.getDisplayName());
-                    FailedTest t = new FailedTest(name, failedTest.isPassed(), errorDetails, stackTrace);
-                    String testYaml = t.toString();
-                    if(printSize <= maxLength && result.tests.size() < maxTests) {
-                        result.tests.add(t);
-                        printSize += testYaml.length();
+            if(maxTests > 0) {
+                int printSize = 0;
+                for (TestResult failedTest : testResult.getFailedTests()) {
+                    if (regressionFilter(failedTest)) {
+                        String stackTrace = showStack ? (escapeHtml ? escapeHtml(failedTest.getErrorStackTrace()) : failedTest.getErrorStackTrace()) : null;
+                        String errorDetails = showMessage ? (escapeHtml ? escapeHtml(failedTest.getErrorDetails()) : failedTest.getErrorDetails()) : null;
+                        String name = String.format("%s.%s", (failedTest instanceof CaseResult) ? ((CaseResult) failedTest).getClassName() : failedTest.getFullName(),
+                                failedTest.getDisplayName());
+                        FailedTest t = new FailedTest(name, failedTest.isPassed(), errorDetails, stackTrace);
+                        String testYaml = t.toString();
+                        if (printSize <= maxLength && result.tests.size() < maxTests) {
+                            result.tests.add(t);
+                            printSize += testYaml.length();
+                        }
                     }
                 }
+                result.otherFailedTests = (failCount > result.tests.size());
+                result.truncatedOutput = (printSize > maxLength);
             }
-            result.otherFailedTests = (failCount > result.tests.size());
-            result.truncatedOutput = (printSize > maxLength);
         }
         return result;
     }
@@ -175,8 +176,9 @@ public class FailedTestsContent extends DataBoundTokenMacro {
         public String toString() {
             StringBuilder builder = new StringBuilder();
             builder.append(this.summary);
-            if(tests.size() == 0) return builder.toString();
-            builder.append(lineBreak);
+            if(tests.size() != 0) {
+                builder.append(lineBreak);
+            }
             for(FailedTest t: tests) {
                 outputTest(builder, lineBreak, t);
             }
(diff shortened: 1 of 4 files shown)


src/test/java/hudson/plugins/emailext/plugins/content/FailedTestsContentTest.java
```diff
@@ -18,9 +18,8 @@
 import java.util.List;
 
 import static java.util.Collections.singletonList;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+
+import static org.junit.Assert.*;
```
slide, line 22: Please maintain the non * imports. Your IDE may have done this for you, I prefer to have the imports listed specifically.

DivyeshChandra: I removed it and retained the non * imports. I dont know how it came up again. Thanks for pointing out.

slide:
Please resolve conflicts now that your other PR has been merged.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
```diff
@@ -106,43 +98,125 @@ else if (result.getRun() != null) {
...
+                for (TestResult failedTest : testResult.getFailedTests()) {
+                    if (regressionFilter(failedTest)) {
+                        String stackTrace = showStack ? (escapeHtml ? escapeHtml(failedTest.getErrorStackTrace()) : failedTest.getErrorStackTrace()) : null;
+                        String errorDetails = showMessage ? (escapeHtml ? escapeHtml(failedTest.getErrorDetails()) : failedTest.getErrorDetails()) : null;
+                        String name = String.format("%s.%s", (failedTest instanceof CaseResult) ? ((CaseResult) failedTest).getClassName() : failedTest.getFullName(),
+                                failedTest.getDisplayName());
+                        FailedTest t = new FailedTest(name, failedTest.isPassed(), errorDetails, stackTrace);
+                        String testYaml = t.toString();
+                        if (printSize <= maxLength && result.tests.size() < maxTests) {
+                            result.tests.add(t);
```
haridsv, line 126: nit: If you decide to take my advise on the approximation of the line count, the `FailedTest` object will no longer be necessary here and it would then be cleaner to have an `addTest()` method in `SummarizedTestResult` to tuck away the above line and the one at 123.

src/test/java/hudson/plugins/emailext/plugins/content/FailedTestsContentTest.java
```diff
@@ -273,4 +274,34 @@ public void testGetContent_withMessage_withStack_htmlEscaped() {
...
+    public void testGetContent_withMessage_withStack_outputYaml() {
+        AbstractTestResultAction<?> testResults = mock( AbstractTestResultAction.class );
+        when( testResults.getFailCount() ).thenReturn( 1 );
+
+        TestResult result = mock( TestResult.class );
+        when( result.isPassed() ).thenReturn( false );
+        when( result.getFullName() ).thenReturn( "hudson.plugins.emailext.ExtendedEmailPublisherTest" );
+        when( result.getDisplayName() ).thenReturn( "Test" );
+        when( result.getErrorDetails() ).thenReturn( "expected:<ABORTED> but was:<COMPLETED> " );
+        when( result.getErrorStackTrace() ).thenReturn( "at org.nexusformat.NexusFile.<clinit>(NexusFile.java:99)" );
```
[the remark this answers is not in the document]
DivyeshChandra: Right now if adding a test case exceeds the `maxLength`, it will continue to add it, but the tests after it will get truncated. So no `FailedTest`'s yaml will ever get truncated abruptly.

src/test/java/hudson/plugins/emailext/plugins/content/FailedTestsContentTest.java
```diff
@@ -273,4 +274,34 @@ public void testGetContent_withMessage_withStack_htmlEscaped() {
...
+        failedTestContent.outputYaml = true;
+        String content = failedTestContent.evaluate(build, listener, failedTestContent.MACRO_NAME);
+        assertEquals(content, "summary: \"1 tests failed.\"\n" +
+                "tests:\n" +
+                "- name: \"hudson.plugins.emailext.ExtendedEmailPublisherTest.Test\"\n" +
+                "  status: \"FAILED\"\n" +
+                "  errorMessage: \"expected:<ABORTED> but was:<COMPLETED> \"\n" +
+                "  stackTrace: \"at org.nexusformat.NexusFile.<clinit>(NexusFile.java:99)\"\n" +
+                "otherFailedTests: false\n" +
+                "truncatedOutput: false\n");
```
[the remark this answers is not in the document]
DivyeshChandra: The order is the order in which the `public` members are declared in the class. So it is predictable.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
DivyeshChandra: Not having a `FailedTest` object will result in a bad Yaml output as I am currently leveraging the yaml library to convert a `List<FailedTest>` to a yaml list structure. The `toString()` can be avoided by just adding up the string lengths of `stackTrace, errorDetails and name` but since we have some boilerplate text along with it in the actual output, i tried to mimic the same behavior in the `toString()` of `FailedTest` so that the estimate is approximately same for each `FailedTest`

src/test/java/hudson/plugins/emailext/plugins/content/FailedTestsContentTest.java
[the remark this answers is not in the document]
haridsv: I am talking about the yaml encoder splitting long lines, not about a test getting split. I am referring to any possible loss of readability and I will get a clarity once you add a real looking stack trace.

> Right now if adding a test case exceeds the maxLength, it will continue to add it

That is an acceptable behavior.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
haridsv: I am not suggesting to get rid off the `FailedTest` class or the use of it in the `SummarizedTestResult` datastructure. I am just referring to the construction of the object here as opposed to abstracting it into a new `addTest` method.

src/test/java/hudson/plugins/emailext/plugins/content/FailedTestsContentTest.java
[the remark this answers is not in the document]
DivyeshChandra: Oh okay. I got it now. I will include a case with proper stacktrace in the unit tests.

DivyeshChandra:
@haridsv Addressed your comments. Please have a look.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
```diff
@@ -143,6 +134,20 @@ private SummarizedTestResult prepareSummarizedTestResult(AbstractTestResultActio
         return result;
     }
 
+    private int addTest(SummarizedTestResult result, int printSize, TestResult failedTest) {
+        String stackTrace = showStack ? (escapeHtml ? escapeHtml(failedTest.getErrorStackTrace()) : failedTest.getErrorStackTrace()) : null;
+        String errorDetails = showMessage ? (escapeHtml ? escapeHtml(failedTest.getErrorDetails()) : failedTest.getErrorDetails()) : null;
+        String name = String.format("%s.%s", (failedTest instanceof CaseResult) ? ((CaseResult) failedTest).getClassName() : failedTest.getFullName(),
+                failedTest.getDisplayName());
+        FailedTest t = new FailedTest(name, failedTest.isPassed(), errorDetails, stackTrace);
+        String testYaml = t.toString();
```
haridsv, line 143: I still feel that this adds a lot of overhead (extra encoding for both YAML and text, the later being a regression in performance). Like I said in my previous comment, I prefer coming up with a heuristic number instead of doing so much more computation to come up with an accurate number. What do you think @slide ?

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
```diff
@@ -73,15 +73,15 @@ public String evaluate(Run<?, ?> run, FilePath workspace, TaskListener listener,
...
-            } catch (Exception e) {
-
-            }
-            return "Bad format";
+        switch (outputFormat) {
+            case "yaml":
+                try {
+                    return result.toYamlString();
+                } catch (Exception e) { }
+                return "Bad format";
```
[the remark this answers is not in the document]
DivyeshChandra: I was actually thinking about how to handle this. I had 2 main thoughts

1) Just print bad format as done, that means that the object cannot be shown as YAML
2) If there is an exception, fall back to the string representation with `tests` as key. 
` "tests: " + result.toString()`

Was not really sure what is more appropriate here.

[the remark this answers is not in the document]
haridsv: > Just print bad format as done, that means that the object cannot be shown as YAML

You could return an encoded string, like "\"Bad format\"" to make it valid YAML, but the issue is that the caller can't blindly use it as a map.

> If there is an exception, fall back to the string representation with tests as key.
"tests: " + result.toString()

Again, same issue of not meeting the contract.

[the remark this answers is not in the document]
slide: You can throw a [MacroEvaluationException ](https://github.com/jenkinsci/token-macro-plugin/blob/master/src/main/java/org/jenkinsci/plugins/tokenmacro/MacroEvaluationException.java )if something is bad.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
slide: I don't think the regression in performance is going to be very noticeable. I think an accurate number is important.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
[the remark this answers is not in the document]
haridsv: @slide What is the consequence of throwing that exception? Would that cause the job to get aborted?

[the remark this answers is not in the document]
slide: Yes, it would be. If there is an error in the macro, I think it should cause an error in the build.

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
```diff
@@ -64,21 +65,22 @@ public boolean acceptsMacroName(String macroName) {
...
         AbstractTestResultAction<?> testResult = run.getAction(AbstractTestResultAction.class);
         SummarizedTestResult result = prepareSummarizedTestResult(testResult);
 
         switch (outputFormat) {
             case "yaml":
                 try {
                     return result.toYamlString();
-                } catch (Exception e) { }
-                return "Bad format";
+                } catch (Exception e) {
```
haridsv, line 81: The only checked exception that can be thrown is `JsonProcessingException` so why catch the generic exception? It adds no benefit in this scenario and is generally a bad practice (anti-pattern).

src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
```diff
@@ -64,21 +65,22 @@ public boolean acceptsMacroName(String macroName) {
...
+                    throw new MacroEvaluationException("Bad format", MACRO_NAME, e.getCause());
```
haridsv, line 82: I have two comments here:
- The message `"Bad format"` is not really helpful at all. I don't think the occurrence of the exception is because of user mistake. On the contrary, this exception indicates either a bug in Jackson or our own intermediate classes. I will leave it to you to come up with a better message.
- I don't see any benefit of using the `e.getCause()` as the cause of the new exception. In fact, this goes on the same lines as the above. If the cause of the exception is a bug in local code (not Jackson), then that will get swallowed and it will then be a nightmare to debug it.

DivyeshChandra: If we catch only `JsonProcessingException` as you suggested and show the cause, it would make sense. Having a cascade of `catch` clauses will help throwing the correct exception i think.

DivyeshChandra:
Hi @slide 
We are waiting on this feature. When is the next release cycle for this plugin ?

PR #209, opened 2020-07-14
[WIP] JCasC validation

alecharp:
Validate the JCasC configuration and "schema" for email-ext.

cc @slide

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.1</version>
+        <version>4.3</version>
     </parent>
 
     <licenses>
@@ -32,6 +32,7 @@
         <hpi.compatibleSinceVersion>2.57.2</hpi.compatibleSinceVersion>
         <concurrency>1</concurrency>
         <argLine>-Dfile.encoding=UTF-8</argLine>
+        <configuration-as-code.version>1.36</configuration-as-code.version>
     </properties>
 
     <developers>
@@ -113,7 +114,7 @@
         <dependency>
             <groupId>org.jsoup</groupId>
             <artifactId>jsoup</artifactId>
-            <version>1.9.1</version>
+            <version>1.11.3</version>
         </dependency>
         <dependency>
             <groupId>net.htmlparser.jericho</groupId>
@@ -123,12 +124,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
-            <version>2.10.0</version>
+            <version>2.10.2</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.10.0</version>
+            <version>2.10.2</version>
         </dependency>
 
         <!-- other plugins -->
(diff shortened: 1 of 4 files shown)


src/test/resources/configuration-as-code.yml
```diff
@@ -0,0 +1,39 @@
...
+        useSsl: true
+        advProperties: "avd-properties"
+    charset: "UTF-8"
+    defaultContentType: "text/html"
+    defaultSubject: "DEFAULT EMAIL SUBJECT"
+    defaultBody: "DEFAULT BODY"
+    emergencyReroute: "emergency-reroute"
+    maxAttachmentSize: 42
+    mailAccount:
+      address: "foo.bar.xyz"
```
slide, line 19: The address field should not be set for the `mailAccount` member.

src/test/resources/configuration-as-code.yml
```diff
@@ -0,0 +1,39 @@
...
+    listId: "list-id"
+    precedenceBulk: true
+    defaultReplyTo: "no-reply@domain.extension"
+    adminRequiredForTemplateTesting: true
+    watchingEnabled: true
+    allowUnregisteredEnabled: true
+    defaultPresendScript: "default-presend-script"
+    defaultPostsendScript: "defaultpostsend-script"
+    defaultClasspath: []
+    defaultTriggerIds: []
```
slide, line 38: FYI, if no trigger ids are specified, then the FailureTrigger will be set (this is for historical reasons). So, if you don't provide anything, make sure the test checks that FailureTrigger is setup in the defaultTriggerIds.

slide:
Perhaps the tests are failing because of the upgrade in jsoup?

alecharp:
@slide probably. I cannot check right now (off for the week but could have access to a computer). I committed before a lunch with family. So it's still in progress.

slide:
It is the jsoup upgrade. I have a fix though.

pom.xml
```diff
@@ -218,6 +219,10 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci</groupId>
+            <artifactId>symbol-annotation</artifactId>
```
slide, line 224: Isn't this part of the structs plugin as well? I already have a dependency on the structs plugin that should provide this.

alecharp: You are absolutely right, I forgot the name of the global lib and only "look" for this one. I'll remove it.

src/test/resources/configuration-as-code.yml
alecharp: I'll have a test with both: some triggers and none. Here it's just because I was in a hurry.. sorry.

src/test/resources/configuration-as-code.yml
alecharp: noted. I'll remove it. Even if it can be configured, not sure of the side effects if it is..

slide: There shouldn't be any side effects that I can think of, just don't want people to look at it as an example.

src/test/resources/configuration-as-code.yml
slide: No worries, I definitely appreciate the help, it was mainly an FYI to make sure you knew before trying to figure out why there was a trigger configured when you setup none :-)

src/test/resources/configuration-as-code.yml
alecharp: > just don't want people to look at it as an example.

I totally agree.
I just put it there, looking at the classes and what was possible to configure there each time.

slide:
I've incorporated most of your changes into my own branch copy, I'll merge them in and then it will be a matter of improving the JCasC tests.

alecharp:
We could have continue to work on this here and merge to your branch when ready. As you have write permission on this repository, you have push permission on my branch as well, when I don't on yours.

slide:
You're right, sorry for that. Was doing it in the background and didn't think about that.

alecharp:
> You're right, sorry for that. Was doing it in the background and didn't think about that.

no problem, it's just I don't want you to feel like I'm dumping all this on you without try to fix anything (even the `TODO`).

slide:
Not at all! I'll still need help for the different JCasC use cases in testing. Shall we close this specific PR?

alecharp:
We can close this, but I'm not sure I can contribute directly to the branch.. I'll need to create a new PR

PR #211, opened 2020-07-28
Validate triggers configuration with casc

alecharp:
cc @slide


slide:
Out of curiosity, what does the "unclassified" mean at the top of the yaml file?

alecharp:
The unclassified is because we didn't create a JCasC category (different from them global management new categories) for email ext. We can add it, I can find an example if you like.

slide:
That's ok, I was just wondering what it meant. Thanks

PR #212, opened 2020-07-28
Fix JENKINS-62444

slide:
This updates email-ext to be compliant with the JCaSC plugin and includes a test for support. Thanks to @alecharp for his help!

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
@@ -267,6 +267,10 @@ public final class ExtendedEmailPublisherDescriptor extends BuildStepDescriptor<
             }
         }
 
+        if(!acc.isValid()) {
+            // what do we want to do here?
+        }
+
         if (acc.getSmtpHost() != null) {
             props.put("mail.smtp.host", acc.getSmtpHost());
         }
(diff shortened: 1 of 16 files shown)


slide:
I have a fix for the build issue, will push it soon

alecharp:
@slide could we cut a release of the plugin with this?

slide:
Yes, I am planning on doing that today.

slide:
The release is out now, had issues with my release machine.

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
```diff
@@ -54,6 +54,7 @@
  * These settings are global configurations
  */
 @Extension
+@Symbol("email-ext")
```
jglick, line 57: This is not backward-compatible unfortunately.

jglick: I would suggest

```java
@Symbol({"email-ext", "extendedEmailPublisher"})
```

if CasC in fact behaves properly: accepting either, but suggesting only the first. @timja @jetersen et al.?

timja: Yes casc works with that, anyone using the previous behaviour will also prompt an administrative warning telling them to update to the new syntax:

e.g. (on my instance which I haven't got around to fixing)

image

batmat: I'm unclear whether this is actually breaking compatibility, or not. IOW, is this supposed to be only a warning, or an instance using the previous values would not start without adapting their yaml (or wait for https://github.com/jenkinsci/email-ext-plugin/pull/271)

timja: it depends on whether someone was using the previous symbol for something or whether any jcasc was possible before without this

PR #213, opened 2020-08-05
Fix JENKINS-63311

slide:
Updated to check for isBlank rather than null for authentication


lukas9393:
```
Sending email to: <e-mail>
MessagingException message: failed to connect, no password specified?
```

Line 325 has a negation, which should not be the case
so it should be StringUtils.isBlank(acc.getSmtpUsername() instead of !StringUtils.isBlank(acc.getSmtpUsername()

slide:
Sorry about that, https://github.com/jenkinsci/email-ext-plugin/pull/215

PR #221, opened 2020-10-17
trim fields to match mailer

darinpope:
Trim fields to match similar changes that were made to mailer.

Also matched descriptions to match mailer.

Finally, I added the same form validator to defaultSuffix that exists in mailer.

Signed-off-by: Darin Pope <darin@planetpope.com>


src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
```diff
@@ -453,6 +454,13 @@ public void setDefaultContentType(String contentType) {
         }
     }
 
+    public FormValidation doCheckDefaultSuffix(@QueryParameter String value) {
+        if (value.matches("@[A-Za-z0-9.\\-]+") || Util.fixEmptyAndTrim(value)==null)
```
proxxz, line 458: Are underscores not allowed in the suffix?

darinpope: according to RFC 1035 (...have as interior characters only letters, digits, and hyphen), underscores are not allowed.

This method was copied from mailer-plugin, so if it should be allowed, it probably should be done in both here and in mailer.

proxxz (approved):

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
proxxz: Thank you for the clarification.  The PR looks great to me!

src/main/java/hudson/plugins/emailext/MailAccount.java
```diff
@@ -23,18 +22,18 @@
 
     @Deprecated
     public MailAccount(JSONObject jo){
-        address = nullify(jo.optString("address", null));
-        smtpHost = nullify(jo.optString("smtpHost", null));
-        smtpPort = nullify(jo.optString("smtpPort", null));
+        address = Util.nullify(jo.optString("address", null));
```
slide, line 25: This constructor is no longer used, it is there only for binary compatibility.

PR #229, opened 2020-10-22
Factor out Jackson version into a property

basil:


jglick (commented):
We cannot use the API plugin here?

basil:
> We cannot use the API plugin here?

I see no reason why we can't use the API plugin. Is that recommended over directly depending on Jackson?

jglick:
As a rule, yes.

basil:
I tried it out and found that the first version of `jackson2-api` that works is 2.11.1, which is a bit recent (June 29, 2020). I want to do at least one release of this plugin with older minimum plugin requirements before I start increasing the minimum Jenkins version and minimum plugin requirements, so I think I'll keep the status quo for now and merge that change after the next release.

PR #236, opened 2020-10-22
Remove javassist from pom.xml

basil:
We don't directly depend on this library, so we can remove any references to it from our `pom.xml`. As far as I can tell, it seems to have been originally added to resolve a `RequireUpperBoundDeps` violation that is no longer present.


jglick (approved):
Yeah I think a newer parent POM rendered this obsolete.

PR #249, opened 2020-10-24
[JENKINS-63846] From address can't be changed

basil:
See [JENKINS-63846](https://issues.jenkins-ci.org/browse/JENKINS-63846). Versions 2.71 and earlier persisted the address for additional accounts but not the default account. Starting in 2.72, the admin address is copied from the Jenkins Location configuration into the default account at the time `ExtendedEmailPublisherDescriptor` is first saved, but this value is not editable in the UI nor does it get updated when the admin address is subsequently updated in the Jenkins Location configuration. Compare the persisted XML [before](https://github.com/jenkinsci/email-ext-plugin/blob/master/src/test/resources/hudson/plugins/emailext/ExtendedEmailPublisherDescriptorTest/persistedConfigurationBeforeJCasC/hudson.plugins.emailext.ExtendedEmailPublisher.xml) and [after](https://github.com/jenkinsci/email-ext-plugin/blob/master/src/test/resources/hudson/plugins/emailext/ExtendedEmailPublisherDescriptorTest/persistedConfigurationBeforeDefaultAddress/hudson.plugins.emailext.ExtendedEmailPublisher.xml). In the second example, you can see that `address` has been persisted for the default account at the time `ExtendedEmailPublisherDescriptor` was saved. Even though [the Jenkins Location configuration](https://github.com/jenkinsci/email-ext-plugin/blob/master/src/test/resources/hudson/plugins/emailext/ExtendedEmailPublisherDescriptorTest/persistedConfigurationBeforeDefaultAddress/jenkins.model.JenkinsLocationConfiguration.xml) was later modified, that change was never propagated back to `ExtendedEmailPublisherDescriptor`.

This change reverts us back to the old (working) design where only additional accounts have a persisted address and the default account has a `null` address, delegating to the Jenkins Location configuration. In order to clear up the bad state that was persisted between 2.72 and 2.77, we unconditionally clear the address for the default account in `ExtendedEmailPublisherDescriptor#readResolve`.

I have added two new tests to verify the correctness of this fix. `ExtendedEmailPublisherTest#testSystemAdminEmailChange` verifies that admin address changes work on the current release, and `ExtendedEmailPublisherDescriptorTest#persistedConfigurationBeforeDefaultAddress` verifies that the persistent state from 2.72 is correctly cleaned up in `ExtendedEmailPublisherDescriptor#readResolve`.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
@@ -192,6 +192,22 @@ public final class ExtendedEmailPublisherDescriptor extends BuildStepDescriptor<
         if(smtpAuthUsername != null) mailAccount.setSmtpUsername(smtpAuthUsername);
         if(smtpAuthPassword != null) mailAccount.setSmtpPassword(smtpAuthPassword);
         if(useSsl) mailAccount.setUseSsl(useSsl);
+
+        /*
+         * Versions 2.71 and earlier correctly left the address unset for the default account,
+         * relying solely on the system admin email address from the Jenkins Location settings for
+         * the default account and using the address specified on the account only for additional
+         * accounts. Versions 2.72 through 2.77 incorrectly set the address for the default account
+         * to the system admin email address from the Jenkins Location settings at the time the
+         * descriptor was first saved without propagating further changes from the Jenkins Location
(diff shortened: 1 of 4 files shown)


timja (approved):

slide (approved):

basil:
Released in [2.78](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.78).

PR #256, opened 2020-11-01
[JENKINS-63522] Debug logs are incomplete when From address is invalid

basil:
See [JENKINS-63522](https://issues.jenkins-ci.org/browse/JENKINS-63522). When debug logs are enabled and the **From** address is invalid (for example, if the **System Admin e-mail address** has not been configured yet in the **Jenkins Location** section of the **Configure System** page and is still set to the default value of `nobody@nowhere`), the outer [`MessagingException`](https://javaee.github.io/javamail/docs/api/javax/mail/MessagingException.html) (typically a [`SendFailedException`](https://javaee.github.io/javamail/docs/api/javax/mail/SendFailedException.html)) is printed to the debug logs but inner exceptions (e.g., [`SMTPAddressFailedException`](https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/SMTPAddressFailedException.html), [`SMTPSenderFailedException`](https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/SMTPSenderFailedException.html), or [`SMTPSendFailedException`](https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/SMTPSendFailedException.html)) are not. The cause is usually found in an inner exception, so it would be helpful to print the inner exceptions to the debug logs when debug logging is enabled.

To test this change, I reproduced the problem and verified that the inner exception is now printed to the debug logs:

```
Successfully created MimeMessage
Sending email to: test@example.com
Error sending to the following INVALID addresses: test@example.com
SendFailedException message: Invalid Addresses
Next SMTPAddressFailedException message: 553 5.1.8 <test@example.com>... Domain of sender address nobody@nowhere does not exist
```

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -508,7 +508,7 @@ public class ExtendedEmailPublisher extends Notifier implements MatrixAggregatab
                                     }
                                     addresses = e.getValidUnsentAddresses();
                                     if (addresses != null && addresses.length > 0) {
-                                        buf = new StringBuilder("Error sending to the following VALID addresses:");
+                                        buf = new StringBuilder("Not sent to the following valid addresses:");
                                         for (Address a : addresses) {
                                             buf.append(' ').append(a);
                                         }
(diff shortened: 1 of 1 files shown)


rvernica:
Note that the log line:
```
Error sending to the following INVALID addresses: test@example.com
```
is still misleading since `test@example.com` is a valid email address!

basil:
Indeed, but this misleading error comes from the [JavaMail](https://javaee.github.io/javamail/) API itself. I do not know of any way around this that does not rely on implementation details of the JavaMail reference implementation, which I am reticent to do.

basil:
I took a closer look at [the API documentation for `SendFailedException`](https://javaee.github.io/javamail/docs/api/javax/mail/SendFailedException.html). Interestingly enough, `getValidUnsentAddresses()` is described as returning "the addresses that are valid but to which this message was not sent", while `getInvalidAddresses()` is described merely as returning "the addresses to which this message could not be sent" without implying anything about their validity (beyond the name of the method). I updated the wording used by the Email Extension plugin to conform more closely to the description of the JavaMail API, removing the implication that the return values of `getInvalidAddresses()` are invalid.

PR #262, opened 2020-11-20
[JENKINS-64141] Restore localhost as default SMTP server

thomasgl-orange:
See [JENKINS-64141](https://issues.jenkins.io/browse/JENKINS-64141).

This is a very minimal change to restore a pre-2.79 behavior: having no SMTP server address (the default setting) means `localhost` will be used, which is often the right choice.
I've kept the `MailAccount.isSmtpServerValid()` method around, because it could still be a place to implement other controls (like checking that the value, if any, looks like an FQDN or IP address), but it now always returns `true`.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -50,7 +50,10 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount>{
     }
 
     public boolean isSmtpServerValid() {
-        return StringUtils.isNotBlank(smtpHost);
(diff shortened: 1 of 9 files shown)


basil:
Thanks for the PR! Based on my reading of the code, the original behavior in Email Extension 2.77 and earlier technically only allowed a blank SMTP server for the default account (not for additional accounts), but I think allowing a blank SMTP server in all cases as in this PR makes more sense.

basil:
Released in [2.80](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.80).

PR #208, opened 2020-05-15
[JENKINS-62290] Make Email Extension Plugin accesible to Jenkins.MANAGE

escoem:
[JENKINS-62290](https://issues.jenkins-ci.org/browse/JENKINS-62290)

This is using the new `Jenkins.MANAGE` permission. It is intended to grant an intermediate access level to global configuration for non admins but more than read users.

The groovy scripts shouldn't be a problem because of the script-security usage.

(I've also updated the the minimum Jenkins core LTS including the `Jenkins.MANAGE` and updated to a newer parent pom).

PR #264, opened 2020-12-08
Update Jenkins baseline, switch to jackson2-api, eliminate reflection, and test against latest LTS

basil:
- Update the Jenkins baseline to 2.222.x, including updating to the latest plugin BOM.
- Depend on the `jackson2-api` plugin rather than directly on Jackson as suggested in [#229 (comment)](https://github.com/jenkinsci/email-ext-plugin/pull/229#pullrequestreview-515207359).
- Now that the baseline is 2.222.x, remove the no-longer-necessary use of reflection introduced in #208.
- Update the `Jenkinsfile` to test against the latest LTS release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,12 +33,12 @@
         <changelist>-SNAPSHOT</changelist>
         <gitHubRepo>jenkinsci/email-ext-plugin</gitHubRepo>
         <java.level>8</java.level>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.222.4</jenkins.version>
         <hpi.compatibleSinceVersion>2.57.2</hpi.compatibleSinceVersion>
         <concurrency>1</concurrency>
         <argLine>-Dfile.encoding=UTF-8</argLine>
-        <configuration-as-code.version>1.36</configuration-as-code.version>
-        <jackson.version>2.11.3</jackson.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -33,12 +33,12 @@
         <changelist>-SNAPSHOT</changelist>
         <gitHubRepo>jenkinsci/email-ext-plugin</gitHubRepo>
         <java.level>8</java.level>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.222.4</jenkins.version>
```
jglick, line 36: FWIW https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ would recommend
```suggestion
        <jenkins.version>2.222.1</jenkins.version>
```
(I have reservations about that recommendation.)

jglick (approved):

pom.xml
basil: This needs to be 2.222.4 because a transitive dependency (`org.jenkins-ci.plugins:credentials:jar:2.3.13`) requires Jenkins 2.222.4 or higher.

jglick: Right, the problem to my mind is lack of consistency in the ecosystem. You can feel free to backdate the version in `credentials` and it should be possible to get a quick release I uess.

PR #308, opened 2021-09-01
Deprecated methods without charset updated

offa:
Updates deprecated calls to methods without explicit charsets.

In addition, usage of deprecated `User.get()` replaced by `User.getById()`. Since `getById()` can return `null`, but `get()` doesn't, the test for `null` is valid now.


basil (approved):
Thank you!

PR #324, opened 2021-10-21
2.277 LTS update

basil:
#323 but based on 2.277 rather than 2.289 and testing against the latest LTS in the `Jenkinsfile`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,9 +8,9 @@ buildPlugin(useContainerAgent: true, configurations: [
   [ platform: 'linux', jdk: '8', jenkins: null ],
 
   // Test the common case (i.e., a recent LTS release) on both Linux and Windows.
-  [ platform: 'linux', jdk: '8', jenkins: '2.263.4' ],
-  [ platform: 'windows', jdk: '8', jenkins: '2.263.4' ],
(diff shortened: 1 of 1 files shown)


batmat:
@basil thanks! this update incidentally is fixing (also) tables-to-divs compatibility and overall PCT compatibility I'm told, would you mind considering a release? Thanks a lot!

basil:
> this update incidentally is fixing (also) tables-to-divs compatibility

I am impressed that tables-to-divs compatibility is fixed in this PR. When I compare the code from before this PR with the code from after this PR, I see no differences that would have fixed tables-to-divs compatibility. I actually see no production code differences at all. This PR only contains changes to `Jenkinsfile`, `pom.xml`, and a test.

batmat:
Yes. I was unclear: there's a PCT failure related to tables-to-divs. I.e. the change in the src/test/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptorTest.java class.
So you're right, no production issue actually here afaict.

basil:
Released in [2.85](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.85).

batmat:
Thank you!

PR #336, opened 2022-01-08
Use detached JavaMail plugin

basil:
Downstream of #335 and jenkinsci/jenkins#6165. Bumps the baseline to one containing jenkinsci/jenkins#6165, thus using the detached JavaMail plugin, but does not upgrade to Jakarta Mail.


basil:
I've marked this as ready for review because the PR is complete, but I'm not in any rush for this to be merged due to the bump in Jenkins baseline. I just wanted to file the PR to close the loop on remaining tasks regarding detaching JavaMail.

PR #361, opened 2022-04-24
fix: JENKINS-62079 - Add contributor() recipient provider which pulls information from ContributorMetadataAction

slide:
This adds a recipient provider which pulls information from the ContributorMetadataAction which is applied to the Run from multiple SCM providers.

https://issues.jenkins.io/browse/JENKINS-62079

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/recipients/ContributorMetadataRecipientProvider.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/recipients/ContributorMetadataRecipientProvider.java
@@ -1,0 +1,82 @@
+package hudson.plugins.emailext.plugins.recipients;
+
+import java.io.PrintStream;
+import java.util.Collections;
+import java.util.Set;
+
+import javax.mail.internet.AddressException;
+import javax.mail.internet.InternetAddress;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jenkinsci.Symbol;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.EnvVars;
+import hudson.Extension;
+import hudson.model.User;
+import hudson.plugins.emailext.ExtendedEmailPublisherContext;
+import hudson.plugins.emailext.ExtendedEmailPublisherDescriptor;
+import hudson.plugins.emailext.Messages;
+import hudson.plugins.emailext.plugins.RecipientProvider;
+import hudson.plugins.emailext.plugins.RecipientProviderDescriptor;
+import jenkins.model.Jenkins;
+import jenkins.scm.api.metadata.ContributorMetadataAction;
+
+public class ContributorMetadataRecipientProvider extends RecipientProvider {
+
+    @DataBoundConstructor
+    public ContributorMetadataRecipientProvider() {
+
+    }
+
+    @Override
+    public void addRecipients(ExtendedEmailPublisherContext context, EnvVars env, Set<InternetAddress> to,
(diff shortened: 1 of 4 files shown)


basil (approved):
Thanks for the PR!

basil:
Released in [2.88](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.88).

slide:
Thanks!

src/main/java/hudson/plugins/emailext/plugins/recipients/ContributorMetadataRecipientProvider.java
```diff
@@ -0,0 +1,115 @@
...
+        }    
+    }
+
+    public User findUser(RecipientProviderUtilities.IDebug debug, String author, String authorEmail) {
+        // first we look for a user based on the user id (author), then if
+        // that fails, we look for one based on the email.
+        User user = null;
+        if(!StringUtils.isBlank(author)) {
+            debug.send("Trying username to get user account from Jenkins");
+            user = User.get(author, false, Collections.emptyMap());
```
KalleOlaviNiemitalo, line 68: GitHubSCMSource.observe stores the GitHub login name of the user as ContributorMetadataAction.getContributor(): <https://github.com/jenkinsci/github-branch-source-plugin/blob/8cb351a1aa74a0c6d888aa5ccecaeb7422731351/src/main/java/org/jenkinsci/plugins/github_branch_source/GitHubSCMSource.java#L2610-L2618>

ContributorMetadataRecipientProvider.findUser then assumes that, if there is a Jenkins user whose login name matches the GitHub user, then they are the same person. This seems it could cause false matches. It might be better to rely primarily on the email address if available, because that one should be globally unique, rather than just unique within GitHub.

It doesn't really matter to me whether this risk is addressed somehow, because I don't think I'll use ContributorMetadataRecipientProvider anyway. Just wanted to point it out.

slide: I thought about that as well and didn't know which people would prefer. I went with this route because I believe most SCM implementations will create a user (unless disabled) and so I thought the username/email would still work for most cases. Looking up a user by the email address will take longer since the whole set of users needs to be looked through to find the user. We can't just use the email address as is in the email-ext because of security checks for sending emails to people who don't have access to the job/run itself. I am open to changing, but I feel like this approach is the most conservative.

KalleOlaviNiemitalo: If this ever becomes a problem, then perhaps it would be best to make the SCM plugins implement a mapping… generate a globally unique ID for the SCM user (so the same login name could mean different people in github.com and GitHub Enterprise even though GitHub Branch Source is used in both cases) and match that to some attribute provided by the authentication realm. That seems it could be quite a complex change.

KalleOlaviNiemitalo: Oh, there's the [User.CanonicalIdResolver extension point](https://www.jenkins.io/doc/developer/extensions/jenkins-core/#user-canonicalidresolver) for this kind of thing, from [JENKINS-11759](<https://issues.jenkins.io/browse/JENKINS-11759> "[JENKINS-11759] Allow single user with multiple usernames, therefore allowing same person to be identified - Jenkins Jira"). So the branch-source plugins could call User.get(String idOrFullName, boolean create, @NonNull Map context) and provide the SCM URI as part of the context. Alas, the extension documentation does not list any implementations that would be able to use that information.

PR #371, opened 2022-06-01
[JENKINS-67701] Add previous() recipient provider 

strangelookingnerd:
Add previous() recipient provider which determines the first culprits / requestor or developers of the previous build(s).
Description of the feature in Jira https://issues.jenkins.io/browse/JENKINS-67701 

The general idea is to find _any_ recipient from a past build so a status change won't go unnoticed.
It would be possible to include this behavior in the culprits / requestor / developers implementation however this would change the current behavior which I guess is not desired.


basil:
@slide Did you have any thoughts about this new recipient provider on a design level? I think the implementation looks fine, so I'm inclined to merge this, unless you think the design of the API could be improved in some way.

src/main/java/hudson/plugins/emailext/plugins/recipients/PreviousRecipientProvider.java
```diff
@@ -0,0 +1,88 @@
...
+                descriptor.debug(logger, format, args);
+            }
+        }
+
+        final Debug debug = new Debug();
+        Run<?, ?> run = context.getRun();
+        Set<User> users = new HashSet<>();
+
+        if (run instanceof RunWithSCM) {
+            users.addAll(((RunWithSCM<?, ?>) run).getCulprits());
```
slide, line 64: This is adding the culprits from the current build, not a previous build. If you specify "previous" as recipient provider, is it expected to get the culprits from the current run?

src/main/java/hudson/plugins/emailext/plugins/recipients/PreviousRecipientProvider.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        Run<?, ?> build = run;
+
+        while (users.isEmpty() && build != null) {
+            users.addAll(RecipientProviderUtilities.getChangeSetAuthors(Collections.singleton(build), debug));
```
slide, line 70: The current build's culprits were added above, should this move to the build.getPreviousCompletedBuild BEFORE the loop so that only previous builds are included?

src/main/resources/hudson/plugins/emailext/Messages.properties
```diff
@@ -24,6 +24,7 @@ CulpritsRecipientProvider.DisplayName=Culprits
 DevelopersRecipientProvider.DisplayName=Developers
 FailingTestSuspectsRecipientProvider.DisplayName=Suspects causing unit tests to begin failing
 FirstFailingBuildSuspectsRecipientProvider.DisplayName=Suspects causing the build to begin failing
+PreviousRecipientProvider.DisplayName=Culprits / Requestor or Developers of the previous build(s)
```
slide, line 27: It seems to me to be an "and" operation, not an or. All of the developers and requester will be added based on the code, so it is not an "or" it is an "and" operation.

slide:
I think previous implies only previous run jobs would be included, but it looks like the current job's developers would be included as well. I think there needs to be a distinction such that previous only looks at previous builds, not at the current build. If you want to separate them out as well, e.g., previousRequestor or previousDevelopers, that might be good as well. Just some thoughts.

src/main/java/hudson/plugins/emailext/plugins/recipients/PreviousRecipientProvider.java
strangelookingnerd: You are absolutely right. I corrected it.

src/main/java/hudson/plugins/emailext/plugins/recipients/PreviousRecipientProvider.java
strangelookingnerd: You are absolutely right. I corrected it.

src/main/resources/hudson/plugins/emailext/Messages.properties
strangelookingnerd: You are absolutely right. I corrected it.

src/main/resources/hudson/plugins/emailext/plugins/recipients/PreviousRecipientProvider/help.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    Sends an email to the the culprits / requestor or developers of the previous build(s).
```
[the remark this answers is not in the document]
strangelookingnerd: You are absolutely right. I corrected it.

strangelookingnerd:
> ... If you want to separate them out as well, e.g., previousRequestor or previousDevelopers, that might be good as well. Just some thoughts.

I thought about that, too. However found that _I_ would not require a separation here and it would only produce lots of duplicated code. If you however feel that it would be good from a design POV I'd be happy to implement it anyway :)

slide (approved):

basil:
Released in [2.89](https://github.com/jenkinsci/email-ext-plugin/releases/tag/email-ext-2.89).

PR #395, opened 2022-09-23
Don't swap spaces for commas.

ajarzyn:
To avoid recipient list friendly name slicing while only one e-mail address is given.

- [ x ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ x ] Ensure that the pull request title represents the desired changelog entry
- [ x ] Please describe what you did
- [ x ] Link to relevant issues in GitHub or Jira
https://issues.jenkins.io/browse/JENKINS-69681
- [ x ] Link to relevant pull requests, esp. upstream and downstream changes
N/A?
- [ O ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
I've edited it in the browser, sorry no environment to run it, and implement tests.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
+++ b/src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
@@ -26,7 +26,7 @@ public class EmailRecipientUtils {
     public static final int TO = 0;
     public static final int CC = 1;
     public static final int BCC = 2;
-    
+
     private static final String EMAIL_LIST_SPLIT_BY = "[ ;,]";
 
     public static Set<InternetAddress> convertRecipientString(String recipientList, EnvVars envVars)
@@ -140,19 +140,15 @@ public class EmailRecipientUtils {
     private static String fixupDelimiters(String input) {
         String[] splitEmailList = input.split(EMAIL_LIST_SPLIT_BY);
         StringBuilder inputBuilder = new StringBuilder();
-        for(String s : splitEmailList)
-        {
-            if(!s.isEmpty())
-            {
+        for (String s : splitEmailList) {
+            if (!s.isEmpty()) {
                 inputBuilder.append(s);
-                if(!s.contains("@")) {
+                if (!s.contains("@")) {
                     inputBuilder.append(" ");
-                }
-                else {
+                } else {
                     inputBuilder.append(",");
                 }
             }
-
         }
         return inputBuilder.toString();
     }
--- a/src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
+++ b/src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
@@ -26,6 +26,8 @@ public class EmailRecipientUtils {
     public static final int TO = 0;
     public static final int CC = 1;
     public static final int BCC = 2;
+    
+    private static final String EMAIL_LIST_SPLIT_BY = "[ ;,]";
 
     public static Set<InternetAddress> convertRecipientString(String recipientList, EnvVars envVars)
             throws AddressException, UnsupportedEncodingException {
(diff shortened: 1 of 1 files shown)


jeffmaury (approved):
/lgtm

basil:
Are we worried about backward compatibility implications here?

ajarzyn:
> Are we worried about backward compatibility implications here?

If peaople were using spaces to separate e-mails adresses. That would be a problem.
But beeing frank I do not know any e-mail client that does that. But I may know only handfull of them.



> This causes tests to fail. Is the goal here to support the format like `First Last <some.email@domain.com>` as a single entity?

Yes indeed. Right now to use it this way we need to add comma at the end of the mail to avoid character swapping. It isn't described anywhere and to find it put I had to look into and analyze the code base.

slide:
We can't break people. Lots of places in Jenkins spaces are used for delimiters in a list. We should just parse this correctly, maybe we can use this: https://docs.oracle.com/javaee/5/api/javax/mail/internet/InternetAddress.html#parse(java.lang.String,%20boolean)

slide:
The help for the address fields generally say to use commas (https://github.com/jenkinsci/email-ext-plugin/blob/c5914a98487eaaf2ee93fcd0059033a3d5806742/src/main/webapp/help/projectConfig/mailType/recipientList.html for instance) , but that doesn't mean people haven't been using spaces. We would need an upgrade path in a readResolve to look for `<` in the fields and update the list to use commas and save back to avoid breakage of people I think. I can look into this soon if the original submitted does not have time.

src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
```diff
@@ -132,10 +132,6 @@ public FormValidation validateFormRecipientList(String recipientList) {
     private static String fixupDelimiters(String input) {
         input = input.trim();
         input = input.replaceAll("\\s+", " ");
-        if(input.contains(" ") && !input.contains(",")) {
-            input = input.replace(" ", ",");
-        }
-
         input = input.replace(';', ',');
         return input;
     }
```
slide, line 137: How about this?

```java
private static String fixupDelimiters(String input) throws AddressException {
        input = input.trim();
        input = input.replaceAll("\\s+", " ");
        input = input.replaceAll(";", ",");
        if(input.contains(" ") && !input.contains(",") && !input.contains("<")) {
            // fix-up spaces to commas
            input = input.replaceAll(" ", ",");
        }
        return input;
    }
```

roboweaver:
I have the same issue, seems reasonable to make the code a bit smarter here so that an address like `Super Admin<super-admin@adobe.com>` wouldn't be treated as two addresses.

My workaround is to not use spaces :)

src/main/java/hudson/plugins/emailext/EmailRecipientUtils.java
ajarzyn: You can't do that check comment with explanation:
https://github.com/jenkinsci/email-ext-plugin/pull/395#issuecomment-1269641355

ajarzyn:
@roboweaver you just reminded me about this PR. I thought it's long closed already.

This proposition solves:
- using spaces, comas, semicolons as delivmeters
- parsing all strings before e-mail as a part of the personals

There's a side defects: `\"name, surname\"`  the `,` semicolon will be removed from the name of the user:
hudson.plugins.emailext.recipients.EmailRecipientUtilsTest.testBackwardsNames
org.junit.ComparisonFailure: expected:<Mouse[,] Mickey> but was:<Mouse[] Mickey>

What can be done?
1. We can split the recipient string following `\"`, next we could ignore one in between `\"` and then glue it back together
2. We can change UT not to include comas in `personals`


I don't really know why this test is failing:
hudson.plugins.emailext.recipients.EmailRecipientUtilsTest.testValidateFormRecipientList_validationShouldPassAListOfGoodEmailAddresses

Is there an easy way to run tests locally?
@jeffmaury @slide

slide:
`mvn clean package` should run all the tests

slide:
Superseded by #562

PR #414, opened 2022-12-30
Adapt test to advanced component change in core

timja:
see https://github.com/jenkinsci/jenkins/pull/7173#issuecomment-1368056507

This is forward and backwards compatible


basil (approved):
Thanks for the PR!

PR #458, opened 2023-06-05
Feature proposal: Email Throttling

tillchen:
The current system does not include any mechanism to limit the number of emails sent within a certain time period. This can lead to a large number of emails being sent in a short time, especially for large projects with frequent builds or many recipients. This is the gap that the proposed Email Throttling feature aims to address.

We also added the throttle bypassing for critical triggers. This is to ensure that high-impact emails are not lost when there is a throttle. The default for email triggers is to not bypass, while certain important triggers override this value to true. 

Note that we have not changed the UI to allow users to configure the email number and time limit. Only admins can change the limits in the code for now. The future plan is to also change the UI to allow the configuration.

### Testing done

We have added automated test cases in `ExtendedEmailPublisherTest.java`.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/EmailThrottler.java
+++ b/src/main/java/hudson/plugins/emailext/EmailThrottler.java
@@ -4,7 +4,7 @@ public class EmailThrottler {
     private static EmailThrottler instance;
     private int emailCount;
     private long lastResetTime;
-    private static final int THROTTLING_LIMIT = 100; // Set your limit
+    public static final int THROTTLING_LIMIT = 100; // Set your limit
     private static final long THROTTLING_PERIOD = 60 * 60 * 1000; // Set your period (in milliseconds)
 
     public EmailThrottler() {
@@ -31,7 +31,7 @@ public class EmailThrottler {
         emailCount++;
     }
 
-    private synchronized void resetEmailCount() {
+    public synchronized void resetEmailCount() {
         emailCount = 0;
         lastResetTime = System.currentTimeMillis();
     }
--- a/src/main/java/hudson/plugins/emailext/EmailThrottler.java
+++ b/src/main/java/hudson/plugins/emailext/EmailThrottler.java
@@ -1,11 +1,10 @@
 package hudson.plugins.emailext;
 
 public class EmailThrottler {
-
-    public static final int THROTTLING_LIMIT = 100; // Set your limit
     private static EmailThrottler instance;
     private int emailCount;
     private long lastResetTime;
+    public static final int THROTTLING_LIMIT = 100; // Set your limit
     private static final long THROTTLING_PERIOD = 60 * 60 * 1000; // Set your period (in milliseconds)
 
     public EmailThrottler() {
--- a/src/main/java/hudson/plugins/emailext/EmailThrottler.java
+++ b/src/main/java/hudson/plugins/emailext/EmailThrottler.java
@@ -1,10 +1,11 @@
 package hudson.plugins.emailext;
 
 public class EmailThrottler {
+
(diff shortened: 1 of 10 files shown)


src/main/java/hudson/plugins/emailext/EmailThrottler.java
```diff
@@ -0,0 +1,38 @@
...
+            instance = new EmailThrottler();
+        }
+        return instance;
+    }
+
+    public synchronized boolean isThrottlingLimitExceeded() {
+        long currentTime = System.currentTimeMillis();
+        if (currentTime - lastResetTime > THROTTLING_PERIOD) {
+            resetEmailCount();
+        }
```
basil, line 26: The returned value of `System.currentTimeMillis` is not guaranteed to be monotonically increasing. Prefer `System.nanoTime` for this purpose instead.

tillchen: Thank you! I have changed it in d894dd2

slide (approved):

omehegan:
Is there any documentation of this feature? I don't see any in the PR and I don't see it mentioned on https://plugins.jenkins.io/email-ext/ Just a line in the release notes.

We have a customer who is impacted by this change and does not really want this feature. It seems like it would have been better to deliver this in an "off by default" state, with an option to turn it on. As written, it will cause unexpected disruption on busy systems, whose admins may not care about the number of emails being sent. They would have to locate this PR and read the code to figure out how to change the throttling settings, which I think can only be done using the Script Console, which is a bit obscure for many people.

tmay-dg:
@tillchen How does someone change the throttling limit? @omehegan did you figure out how to change it? I agree that this is disruptive and it's unclear how to change the behavior.

duemir:
Since those static fields are `final` there is no easy way to change them from [Jenkins Script Console](https://www.jenkins.io/doc/book/managing/script-console/).
https://gist.github.com/pditommaso/263721865d84dee6ebaf documents a potential trick, but I cannot make it work in my test environment. Something like:

```
import java.lang.reflect.Field
import java.lang.reflect.Modifier

println("THROTTLING_LIMIT Before: ${hudson.plugins.emailext.EmailThrottler.THROTTLING_LIMIT}")
def throttlingLimitField = hudson.plugins.emailext.EmailThrottler.class.getDeclaredField("THROTTLING_LIMIT")

// Remove `final` using reflection to be able to modify the value
Field modifiersField = Field.class.getDeclaredField("modifiers")
modifiersField.setAccessible(true)
modifiersField.setInt(throttlingLimitField, throttlingLimitField.getModifiers() & ~Modifier.FINAL)

// Update the THROTTLING_LIMIT now that it should no longer be final
throttlingLimitField.setInt(null, 1000)
println("THROTTLING_LIMIT After: ${hudson.plugins.emailext.EmailThrottler.THROTTLING_LIMIT}")
```

tmay-dg:
If anyone is curious, we did find a way to override this behavior from the Jenkins Script Console, by overriding the `isThrottlingLimitExceeded` method:
```
import hudson.plugins.emailext.EmailThrottler;
EmailThrottler.instance = new EmailThrottler() {
  public synchronized boolean isThrottlingLimitExceeded() {
    return false;
  }
}
```
Sadly this only works until Jenkins is restarted, so you'd want to put this in to a groovy script and make use of https://www.jenkins.io/doc/book/managing/groovy-hook-scripts/#post-initialization-script-init-hook

Dohbedoh:
Create https://issues.jenkins.io/browse/JENKINS-73740 to make sure this is made configurable eventually.

PR #481, opened 2023-08-30
[JENKINS-71925] Deprecation of jobs modifying the 'from' email field

rkivisto:
[JENKINS-71925](https://issues.jenkins.io/browse/JENKINS-71925)

Re-implementing this feature after discussion with @basil in https://github.com/jenkinsci/email-ext-plugin/pull/479

The email-ext plugin previously allowed users configuring jobs on a controller to modify the 'from' attribute at a job level (using Pipeline `from:` , or freestyle `Project From`).

Some companies have internal policies which do not allow this, and it can be unexpected when an email 'from' can be controlled by any user who has the `Job/Configure` permission or the ability to commit to a repo which contains a Jenkinsfile.

After this change, the 'from' field of emails sent by the email-ext plugin should always use the `from` address defined at `Manage Jenkins` -> `System` -> `System Admin e-mail address`, and not be configurable at a job-level.

Users that would like to go back to the previous behaviour can add the system property as a startup argument: `-Dhudson.plugins.emailext.enable-job-from=true`.

A message is emitted to the controller logs when the 'from'/'project from' field is used by a job, at a `WARNING` level if the 'from' address was ignored, and 'INFO' level if the system property is set (to remind about the deprecation).

### Testing done

1 . Start the instance with the system property to continue to allow 'from' feature to be available:

```
mvn hpi:run -Dhudson.plugins.emailext.enable-job-from=true
```

2 . Configure the email-ext plugin with my email provider
3 . Set my email address at `Manage Jenkins` -> `System` -> `System Admin e-mail address`
4 . Create a Pipeline job which uses the `from` field with a fake address (sent to myself)

```
emailext body: 'test', subject: 'test-using-from', to: 'me@example.com', from: 'user2@example.com'
```

5 . When this job is run, I see `X-Google-Original-From:` in the email content, and the following message is logged (this message is not emitted if `from:` is not used):

```
INFO    h.p.e.ExtendedEmailPublisher#getFromAddress: JENKINS-71925: Job has `Project From` configured, which is deprecated
```

6 . Create a Freestyle job, adding a `Post-build Actions` of type `Editable Email Notification`, and filling in the `Project From` field, then saving the job. When the job is saved, and the `Project From` field is is not blank, the following message is logged:

```
INFO    h.p.e.ExtendedEmailPublisher#<init>: JENKINS-71925: Freestyle job was configured with `Project From` which is deprecated
```

7 . When this freestyle job runs, I see `X-Google-Original-From:` in the email, and the following message is logged:

```
INFO    h.p.e.ExtendedEmailPublisher#getFromAddress: JENKINS-71925: Job has `Project From` configured, which is deprecated
```

8 . Stop the controller and start it without the new system property (so the feature is disabled by default):

```
mvn hpi:run
```

9 . Run the Pipeline job from step 4, the email will not contain `X-Google-Original-From:`, and the following warning is logged:

```
WARNING h.p.e.ExtendedEmailPublisher#getFromAddress: JENKINS-71925: Job has `Project From` configured, from value was ignored because it is disabled by default. Enable by setting system property 'hudson.plugins.emailext.enable-job-from' to true
```

10 . Run the freestyle job from step 6, the email will not contain `X-Google-Original-From:`, and the following warning is logged:

```
WARNING h.p.e.ExtendedEmailPublisher#getFromAddress: JENKINS-71925: Job has `Project From` configured, from value was ignored because it is disabled by default. Enable by setting system property 'hudson.plugins.emailext.enable-job-from' to true
```

11 . I also tested Pipeline and freestyle jobs which do not set the `from` field, they work correctly, and do not emit any messages in the controller logs (the email comes from the `System Admin e-mail address`)

12 . Configure the freestyle job from step 6, notice the `Project From` field no longer shows up in the UI, and when the job is saved, the config is saved without `Project From`.

13 . Stop the instance again, and add the system property:

```
mvn hpi:run -Dhudson.plugins.emailext.enable-job-from=true
```

14 . Ensure the 'from' feature still works for freestyle and Pipeline jobs, but the job which was saved in step 12 has the `Project From` field empty

### Submitter checklist

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -310,6 +310,18 @@ Class groovyClass = new GroovyClassLoader(getClass().getClassLoader()).parseClas
 GroovyObject jiveFormatter = (GroovyObject) groovyClass.newInstance();
 ----
 
+=== JENKINS-71925: Deprecation of jobs modifying the 'from' email field
+
+In plugin version `2.100` and older, users could configure jobs on a controller to modify the 'from' attribute at a job level (using Pipeline `from:` , or freestyle `Project From`).
+
+Some companies have internal policies which do not allow this, and it can be unexpected when an email 'from' can be controlled by any user who has the `Job/Configure` permission or the ability to commit to a repo which contains a Jenkinsfile.
+
+The 'from' field of emails now always uses the `from` address defined at `Manage Jenkins` -> `System` -> `System Admin e-mail address`, and is not configurable at a job-level.
+
+Users that would like to go back to the previous behaviour can add the system property as a startup argument: `-Dhudson.plugins.emailext.enable-job-from=true`.
+
+A message is emitted to the controller logs when the 'from'/'project from' field is used by a job, at a `WARNING` level if the 'from' address was ignored, and 'INFO' level if the system property is set (to remind about the deprecation).
+
 == Related plugins
 
 * https://plugins.jenkins.io/email-ext-recipients-column/[Email Ext Recipients Column plugin]
--- a/README.adoc
+++ b/README.adoc
@@ -318,7 +318,7 @@ Some companies have internal policies which do not allow this, and it can be une
 
 The 'from' field of emails now always uses the `from` address defined at `Manage Jenkins` -> `System` -> `System Admin e-mail address`, and is not configurable at a job-level.
 
-Users that would like to go back to the previous behaviour can add the system property as a startup argument: `-Dhudson.plugins.emailext.enable-job-from=true`.
+Users that would like to go back to the previous behaviour can add the system property as a startup argument: `-Dhudson.plugins.emailext.ExtendedEmailPublisher.enableFrom=true`.
 
 A message is emitted to the controller logs when the 'from'/'project from' field is used by a job, at a `WARNING` level if the 'from' address was ignored, and 'INFO' level if the system property is set (to remind about the deprecation).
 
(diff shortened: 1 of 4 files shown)


basil:
I took a closer look at this today and pushed a few cleanup commits. While reading the code more closely, I discovered that there is an Additional Accounts feature in Email Extension (it's buried beneath the Advanced button) which allows an administrator (not end user!) to specify a custom Admin Account Address and SMTP credentials. This does actually seem useful (and secure!) and I don't think we would want to break this feature. But to use this feature (which is extremely poorly documented), one has to set the Project From to the value of one of the Admin Account Addresses from one of the Additional Accounts. And this PR would break this use case.

Another way of looking at this is that the problem is not the Project From field per se, but rather that it can be set to any arbitrary value, rather than blank (which selects the Jenkins-wide email address) or the value from one of the Additional Accounts (which a system administrator must approve). When `ENABLE_FROM` is true I think we should print a message that the provided From address does not match either the default account or any additional accounts (but use it anyway), and when `ENABLE_FROM` is false I think we should print the same message but treat it as if it were blank (i.e., use the default account).

basil:
From a user experience point of view, it might be a good idea to show a dropdown in Project From rather than a plain text box:

- [Default account] jenkins@example.com
- [Additional Account 1] foo@example.com
- [Additional Account 2] bar@example.com

And make the user select one of them. However, I'm not sure if this would be acceptable from a security perspective, as it would expose the from addresses of the additional accounts to someone editing the job. Maybe we could only show it if the user editing the job is an administrator. In any case, this is not mandatory for this PR, but an optional improvement.

rkivisto:
Seems a redesign of this idea would be required.
Maybe the simpler system property idea (#479) would be a good short term solution instead?

PR #518, opened 2024-05-10
Do not submit an empty pre/post-send script for approval

jglick:
Addresses what seems to have been a regression from 04efc23: every time you save the global config page, a warning pops up in **Manage** that you have a script to approve…but it is empty. Note that the method call here is deprecated as of https://github.com/jenkinsci/script-security-plugin/commit/f4c0bb9b58e105b4fc6b62be0f7f2daa46178190 but I am not touching that.


Vlatombe (approved):

jglick:
I would appreciate a release when you get a moment BTW!

jglick:
(That would be automatic if an appropriate label were applied prior to merge.)

jglick:
@slide could this be released please?

slide:
> @slide could this be released please?

In progress now, sorry, I should have done that sooner.

jglick:
https://github.com/jenkinsci/email-ext-plugin/releases/tag/1814.v404722f34263 thanks! If you apply labels like `bug` prior to merge, then the release should happen without human intervention. (There is no open RFE to create a check that blocks merge until _some_ known label is applied.)

PR #569, opened 2024-12-04
[JENKINS-74891] Extract inline JavaScript from `EmailExtTemplateAction/index.groovy`

yaroslavafenkin:
https://issues.jenkins.io/browse/JENKINS-74891

There's still a problem with the mechanism that renders the template. It's rendered inside an `iframe` element. Default CSP header that's set by [Content Security Policy](https://plugins.jenkins.io/csp/) does not define `frame-src`, hence `default-src` is used, which only allows `'self'`. This means that `data:` src for `iframe` is not allowed, which prohibits templates from being rendered in restrictive mode. This is also be demonstrated on "After the change" video below.

Depending on how wanted this feature is it can either be fixed or deprecated with a notice that it doesn't work in CSP restrictive mode. 
Alternatively CSP plugin could define `frame-src` policy that we'd consider safe (probably a question to @jenkinsci/core-security-review what sources would be safe). I have doubts that `data:` will make it into the list of safe sources.

### Testing done

[Before the change](https://www.youtube.com/watch?v=TI5BiUTMsyY)
[After the change](https://www.youtube.com/watch?v=MUt2rLCHQ6I)

### Submitter checklist

### DIFF
--- a/src/main/resources/hudson/plugins/emailext/EmailExtTemplateAction/index.groovy
+++ b/src/main/resources/hudson/plugins/emailext/EmailExtTemplateAction/index.groovy
@@ -14,25 +14,11 @@ l.layout {
     st.include(it: my.project, page: "sidepanel")
     l.main_panel {
         st.bind(var: "templateTester", value: my)
-        script """function onSubmit() {
-                var templateFile = document.getElementById('template_file_name').value;
-                var buildId = document.getElementById('template_build').value;
-                templateTester.renderTemplate(templateFile,buildId, function(t) {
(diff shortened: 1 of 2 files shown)


slide:
Could the iframe be replaced with an `<object>` tag? I am not a web developer by any means, but it looked from googling that it might be possible?

slide:
I did a quick check with using <object> instead of <iframe> and I see a template rendered correctly, but it does not bypass the issue with CSP as it sets .data on the object tag similar to the src for the iframe. Is there a way to render HTML with CSP enforcement?

slide:
I think all of that is outside the scope of this PR, so I'm happy to merge and look at options for full compliance later.

slide (approved):

PR #1465, opened 2026-02-24
[JENKINS-75800] Allow email attachments to have inline content-disposition

harshavardhanreddy06:
This PR implements support for inline attachments as requested in https://github.com/jenkinsci/email-ext-plugin/issues/1438.

### Changes:
- Added `inlineAttachmentsPattern` field to `EmailType`, `ExtendedEmailPublisher`, and `EmailExtStep`.
- - Updated `AttachmentUtils` to support `Part.INLINE` disposition for attachments matching the inline pattern.
- - Added "Inline Attachments" UI fields in Project configuration, Trigger configuration, and Pipeline step.
- - Standardized `AttachmentUtils.getAttachments` to take a boolean flag for inline disposition.
This allows users to specify patterns for files that should be treated as inline content (e.g., images for CID embedding), while keeping existing attachment behavior intact.


Philodoescode:
I guess simply setting Content-Disposition: inline is not enough, If you add an inline attachment to a standard multipart/mixed email, Teams will display it inline AND in the attachment drawer. To properly hide an inline image from the attachment drawer, it must be nested inside a multipart/related part alongside the HTML body.
The email structure needs to look like this:
multipart/mixed:
	multipart/related:
		text/html (The email body)
		image/png (Content-Disposition: inline, Content-ID: <banner.png>)
	application/zip (Content-Disposition: attachment)

ArpanC6:
Hi @harshavardhanreddy06! The reviewer feedback from @slide has been addressed — good work restoring the Javadoc and switching to @DataBoundSetter.
@Philodoescode raises a valid point. Setting Content-Disposition: inline alone is not sufficient for Microsoft Teams. The correct MIME structure requires multipart/related wrapping the HTML body and inline attachments together. PR #1499 takes this approach.
Two competing solutions now exist for the same issue:

This PR (#1465): adds inlineAttachmentsPattern field with Content-Disposition: inline
PR #1499: restructures MIME to use multipart/related

These approaches could potentially be combined — use inlineAttachmentsPattern to identify which attachments should be inline, then place them inside a multipart/related block as #1499 suggests.
Worth discussing with @slide which direction to take.

slide (changes requested):

slide:
This is a solid fix — `multipart/related` is the correct RFC 2387 structure for inline attachments, and the four integration tests that verify the actual MIME structure are exactly the right kind of coverage for a change like this.

### Plain text content type with inline attachments silently drops them

The inline attachment path is gated on `messageContentType.startsWith("text/html")`. If a user configures plain text content with an inline attachments pattern, the attachments are silently dropped with no log message or warning. At minimum a warning should be logged explaining why inline attachments are being skipped, so users can understand the misconfiguration.

### `testMultipleInlineAttachments` has a weak Content-ID assertion

```java
assertTrue(cid1.equals("<blank.gif>") || cid1.equals("<test.pdf>"));
assertTrue(cid2.equals("<blank.gif>") || cid2.equals("<test.pdf>"));
```

This allows both `cid1` and `cid2` to be `<blank.gif>` — it doesn't actually verify both files are present. Use a `Set` instead:

```java
Set<String> cids = Set.of(cid1, cid2);
assertTrue(cids.contains("<blank.gif>") && cids.contains("<test.pdf>"));
```

### The `doBoth` + inline attachments path has no test

The `doBoth` branch (content type `"both"`) wraps `htmlBlock` in a `relatedWrapper` — the most structurally complex path in the new code — and has no test coverage. A test using `contentType = "both"` with an inline attachment would be valuable to confirm the `alternative → related → (html, inline)` structure is correct.

### Removal of `inlineAttachmentsPattern` handling from `createMail` should be called out

The original code called `attachInline` on the top-level multipart directly from `createMail`. The new code moves this into `addContent`. This is an intentional structural change but it's not mentioned in the PR description — worth adding a note so reviewers understand why the placement changed.

PR #1467, opened 2026-02-25
Account for OAuth2 in credential validation

subwaycookiecrunch:
[doCheckCredentialsId](cci:1://file:///c:/Users/raj97/.gemini/antigravity/scratch/email-ext-plugin/src/main/java/hudson/plugins/emailext/MailAccount.java:136:8-183:9) doesn't take `useOAuth2` as a parameter so it shows the same TLS/SSL warning regardless of whether OAuth2 is enabled or not.

This adds `useOAuth2` as a `@QueryParameter`. When it's checked without TLS/SSL, the validator now shows a softer warning instead of the FIPS error, since the password is a token not a real password.

### Testing done

Checked that the field name in [config.groovy](cci:7://file:///c:/Users/raj97/.gemini/antigravity/scratch/email-ext-plugin/src/main/resources/hudson/plugins/emailext/MailAccount/config.groovy:0:0-0:0) lines up with the new parameter. The FIPS tests in `MailAccountFIPSTest` still pass since that code path is unchanged.

### Submitter checklist

Related: #1420

### DIFF
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -68,7 +68,8 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
     }
 
     public boolean isSecureAuthWhenFIPS() {
-        // when in FIPS mode if we are using authentication we must also use TLS or SSL to protect the password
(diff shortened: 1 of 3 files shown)


slide:
I think I will hold off on this until after the OAuth flow is reworked during GSoC. Please refile a PR after GSoC if these changes are still needed.

PR #1468, opened 2026-02-26
Set inline content-disposition for attachments with Content-ID

subwaycookiecrunch:
When an attachment has a `Content-ID` (e.g. for ``), it shows up both inline in the email body and again as a file attachment at the bottom. This is especially annoying in Microsoft Teams where you get the image twice.

The fix is just setting `Content-Disposition: inline` on the MIME part via `Part.INLINE`, so mail clients know not to also show it as a separate download.

Fixes [JENKINS-75800].

### Testing done

Compile + existing tests pass. The change itself is one `setDisposition` call on a `MimeBodyPart` — no branching logic to test, it just sets a MIME header.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
+++ b/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
@@ -15,6 +15,7 @@ import jakarta.activation.DataSource;
 import jakarta.activation.FileTypeMap;
 import jakarta.mail.MessagingException;
 import jakarta.mail.Multipart;
+import jakarta.mail.Part;
 import jakarta.mail.internet.MimeBodyPart;
 import jakarta.mail.internet.MimeUtility;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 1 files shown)


ArpanC6:
Hi @subwaycookiecrunch! This is a minimal and clean fix. All CI checks are passing — great work!
One observation: there's a competing PR #1499 that takes a different approach — restructuring the MIME tree to use multipart/related for inline attachments.
Your approach (setting Content-Disposition: inline) is simpler, but some email clients (especially Outlook) may still show duplicates because they rely on multipart/related structure, not just the Content-Disposition header.
PR #1499's approach is more RFC-compliant but breaks existing tests. Your approach is minimal and safe.
Worth noting for @slide to decide which approach to merge — both fix JENKINS-75800 but at different levels.

slide (changes requested):
Not sure what this is trying to achieve. We don't want all attachments as inline.

subwaycookiecrunch:
@slide yeah fair point, I should've explained this better. The thing is `setContentID` is already being called on every attachment (line 171, that's not my change), so mail clients already see a Content-ID on them. But the disposition is still `attachment`, so you end up with the image showing inline *and* as a separate download at the bottom. This was driving people crazy in Teams especially.

All I did was add `setDisposition(Part.INLINE)` to match the Content-ID that's already being set. But yeah setting everything inline isn't ideal , I can guard it so it only kicks in when the HTML body actually has a matching `cid:` reference. Or if you'd rather go with the `multipart/related` approach in #1499 I'm fine closing this. Let me know what works.

slide:
Yeah, I think we'll go a different direction on this. Thanks for the contribution though!

PR #1477, opened 2026-03-06
[JENKINS-75238] Fix truncated Editable Email Notification configuration UI

woeter69:
Fixes [JENKINS-75238] / #1431

## Problem

The "Editable Email Notification" configuration block fails to render fully in modern Jenkins versions. When users add it as a post-build action, the "Advanced Settings..." button, Triggers, and other settings below the "Content Token Reference" entry are invisible.

## Root Cause

In both config.groovy and global.groovy, the "Content Token Reference" form entry used field: "tokens":

    f.entry(title: _("Content Token Reference"), field: "tokens")

The field attribute instructs Stapler to data-bind to a tokens property on the model object. No such property exists on ExtendedEmailPublisher or its descriptor. The field was only used as a shortcut to locate help-tokens.groovy. In older Jenkins this was silently ignored, but in current versions it throws a groovy.lang.MissingPropertyException that breaks the rendering of the entire configuration block.

## Fix

Replaced "tokens" with help: descriptor.getHelpFile("tokens") in both files. This resolves the help file URL through the standard Jenkins descriptor mechanism without triggering any data-binding on a nonexistent property.

### Testing

All existing tests pass (ExtendedEmailPublisherDescriptorTest: 19/19, 0 failures).


ArpanC6:
Hi @Woeter69! The fix is minimal and correct — just 2 lines changed across config.groovy and global.groovy.
The change from field: "tokens" to help: descriptor.getHelpFile("tokens") is exactly the right Stapler idiom. The field attribute triggers data-binding lookup on the model object, while help simply resolves the help file URL — no data-binding involved.
All 17 CI checks passing. This is a clean, surgical fix for a real regression.

slide:
Clean fix — `field` was never the right attribute for rendering a help link, and switching to `help: descriptor.getHelpFile("tokens")` is the correct idiomatic Jenkins approach. Applied consistently in both `config.groovy` and `global.groovy`.

Could you add a screenshot to the PR showing the Content Token Reference help link rendering correctly in the UI? The code change is straightforward but a quick visual confirmation would help. A before and after would be even better!

Faheem12005:
### After:
Post build action:
image
Advanced Settings dropdown:
image

The UI does render correctly after the changes
However, I wasn't able to reproduce the original issue locally. With the pre-fix version, the Advanced Settings dropdown still rendered correctly and I was able to interact with it. If someone is able to double check would be great

woeter69:
native_before
Before (Advanced Settings button is not clickable here)
after_closed
after_opened
After


The 'Before' screenshot was reproduced by manually throwing MissingPropertyException  to copy the strict JEXL rendering failure possibly seen in production. 
I was not able to create this bug in local development environment because it suppresses strict JEXL errors.

slide:
Thanks!

PR #1480, opened 2026-03-07
[JENKINS-65558] Added Configuration as Code support for email templates

### LINKED ISSUE #1356: [JENKINS-65558] Add CasC support for /email-templates/
<p>Email templates<span class="error">&#91;1&#93;</span> are expected to be added to JENKINS_HOME. Provide a way to manage them through CasC instead.</p>

<p><span class="error">&#91;1&#93;</span> <a href="https://plugins.jenkins.io/email-ext/#jelly-script-content-templates" class="external-link" target="_blank" rel="nofollow noopener">https://plugins.jenkins.io/email-ext/#jelly-script-content-templates</a></p>

---
<details><summary><i>Originally reported by  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=olivergondza">olivergondza</a>, imported from: <a class="original-jira-link" href="https://issues.jenkins.io/browse/JENKINS-65558" target="_blank">Add CasC support for /email-templates/</a></i></summary>
<i><ul>
<li><b>status</b>: Open
<li><b>priority</b>: Major
<li><b>votes</b>: 0
<li><b>watchers</b>: 3
</ul></i>

<pre>
<p>Email templates<span class="error">&#91;1&#93;</span> are expected to be added to JENKINS_HOME. Provide a way to manage them through CasC instead.</p>

<p><span class="error">&#91;1&#93;</span> <a href="https://plugins.jenkins.io/email-ext/#jelly-script-content-templates" class="external-link" target="_blank" rel="nofollow noopener">https://plugins.jenkins.io/email-ext/#jelly-script-content-templates</a></p></pre>
</details>
</details>

woeter69:
Resolves #1356 / [JENKINS-65558]

## Problem

Custom email templates used by the Email Extension Plugin (Groovy, Jelly, etc.) must be manually placed in the `$JENKINS_HOME/email-templates/` directory on the Jenkins master. There was no way to manage these templates programmatically using Jenkins Configuration as Code (JCasC), which complicates automated provisioning and infrastructure-as-code deployments.

## Solution

This PR adds support for provisioning email templates directly via JCasC YAML configuration. 

Templates defined within the `email-ext` unclassified configuration are automatically validated and written to the `$JENKINS_HOME/email-templates/` directory during Jenkins startup or when CasC configuration is reloaded.

### Security Implementation
Strict validation is applied to prevent any malicious file writing:
- **Filename constraints:** Rejects names containing path traversal characters (`/`, `\`, `..`), null bytes, or spaces.
- **Extension whitelist:** Enforces that template files must end with `.groovy`, `.jelly`, or `.template`.
- **Path Checking:** A canonical path check ensures the final resolved template file strictly remains within the bounds of the `email-templates/` directory.

## Usage (JCasC)

```yaml
unclassified:
  email-ext:
    emailTemplates:
      - name: "custom-template.groovy"
        content: |
          def subject = "Build Notification"
          def body = "Build ${PROJECT_NAME} finished with status ${BUILD_STATUS}."
          return body
      - name: "html-alert.jelly"
        content: |
          <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define">
            <h1>Build Failed</h1>
          </j:jelly>

### CODE REVIEW

**Comment by slide** [comment]
@jenkinsci/core-security-review Can you look this over and tell me if you see any security issues with this? I think it should be ok since admins would be the ones using JCasC to provision Jenkins, but I just want to make sure there is no issue.

**Review by daniel-beck — CHANGES_REQUESTED** [review]
Some questionable decisions. I would not be surprised at all to find out that this was vibecoded.

https://github.com/Woeter69/email-ext-plugin/blob/7e820bc4a2157202a7d5f46b01b92592ccacfb7b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java#L942 means that this would allow Overall/Manage users to configure by submitting crafted JSON form submissions. This would be a big security issue.

I would not accept it in its current form, but the general idea is reasonable.

TBH my suggestion would be to move away from file based templates. Also a nice opportunity to move away from the current set of (weird) templates (who's Larry!?).

**Comment by woeter69** [comment]
Yes, I'll start working on the things mentioned ASAP.

**Comment by ArpanC6** [comment]
Hi @Woeter69! After reviewing daniel-beck's feedback, here's a summary of what needs to be addressed:
1. Security issue (most critical): Remove @DataBoundSetter from setEmailTemplates() or restrict it to ADMINISTER permission check, since currently any Overall/Manage user could exploit this via crafted JSON to write arbitrary files.
2. Windows path fix: Replace getCanonicalPath() with java.nio.Path#startsWith:
javaPath resolvedTemplate = templateFile.toPath().toRealPath();
Path resolvedDir = templatesDir.toPath().toRealPath();
if (!resolvedTemplate.startsWith(resolvedDir)) { ... }
3. Template cleanup: In setEmailTemplates(), before writing new templates, collect existing template filenames and delete ones not present in the new list.
4. Simplify validation: The manual checks in EmailTemplate.java (lines 46-61) are redundant if the regex already covers them — remove the manual checks and rely on the regex.
Hope this helps with the fixes!

**Comment by woeter69** [comment]
i've done some changes, please review my changes and provide me with some feedback, Thanks!

**Comment by ArpanC6** [comment]
Hi @Woeter69! Great work addressing the feedback. Here's what I can see has been fixed:
1)  Security fix: configure() method now checks Jenkins.ADMINISTER permission before processing emailTemplates
2) Windows path fix: getCanonicalPath() replaced with nio.Path#normalize().toAbsolutePath() and startsWith() — correct approach
3) Template cleanup: Old templates are deleted before writing new ones — accumulation issue resolved
4) Validation simplified: EmailTemplate.java now relies solely on SAFE_NAME_PATTERN regex — manual checks removed
5) Tests added: shouldProvisionEmailTemplates() and shouldRejectInvalidTemplateNames() cover the main scenarios
One remaining concern: The @DataBoundSetter on setEmailTemplates() is still present. daniel-beck's original concern was that this allows form submission bypass — even with the configure() permission check, a crafted JSON POST to the form endpoint could still call the setter directly. Consider removing @DataBoundSetter and handling this only through configure().
Overall this is significantly improved!

**Comment by woeter69** [comment]
@slide i've made some changes after the review, i'm looking forward to some feedback, thank you.

**Comment by slide** [comment]
I'll review soon

**Comment by slide** [comment]
Well-thought-out feature — the `SAFE_NAME_PATTERN` regex, the canonical path check, and the permission guard all show careful security thinking. The JCasC test with a real YAML fixture is the right approach for end-to-end verification.

### `setEmailTemplates` silently deletes user-managed templates on every reload

The cleanup block deletes all `.groovy`, `.jelly`, and `.template` files from `email-templates/` before writing the new set. This means any templates manually placed there by users — not managed via CasC — will be silently deleted every time JCasC reloads, including on Jenkins restart. This is a significant and potentially surprising data loss scenario.

Consider tracking which files were provisioned by CasC (e.g. a manifest file or a naming convention) so that only those are removed on reload, leaving user-managed templates untouched.

### Permission check in `configure()` is in the wrong place

```java
if (formData.has("emailTemplates")) {
    Jenkins.get().checkPermission(Jenkins.ADMINISTER);
}
```

This relies on the UI not sending `emailTemplates` in the form data for non-admin users, which is fragile. The permission check would be more robust inside `setEmailTemplates()` itself, so it applies regardless of how the setter is called.

### No test for the cleanup behaviour on reload

The tests verify that templates are written correctly and that invalid names are rejected. But there's no test for what happens when a template is removed from the CasC YAML on the next reload — which is the most surprising behaviour in the implementation. A test that provisions templates, then reloads with a reduced set, and confirms the removed template file is gone (but any user-managed files are preserved) would be valuable.

woeter69:
I've addressed all 3 mentioned points:

### 1. Preserved user-managed templates on reload

`setEmailTemplates()` from now on writes a `.casc-managed` manifest file to `email-templates/` listing the filenames it provisioned. On the next reload, only files present in that manifest are deleted before writing the new set. User-managed templates will not be touched at all.

### 2. Permission check moved into `setEmailTemplates()`

The `Jenkins.ADMINISTER` check is now the first line of the setter itself. It's enforced no matter how the method is actually called (UI, CasC, or API directly). The check in `configure()` was removed.

### 3. Added a test for cleanup behaviour on reload

New test `shouldCleanUpRemovedTemplatesButPreserveUserManaged`:

1. Loads the full template YAML (groovy + jelly)
2. Places a user-managed `user-custom.groovy` in the directory
3. Reloads with a reduced YAML (groovy only)
4. Asserts the removed jelly template is deleted
5. Asserts the user-managed template is preserved with intact content
6. Asserts the manifest is updated to reflect only the remaining CasC template

---

**Status:** All 33 descriptor tests pass (`mvn spotless clean`).

**Files changed:**

* `ExtendedEmailPublisherDescriptor.java` - manifest-based cleanup + permission guard
* `ExtendedEmailPublisherDescriptorJCasCTest.java` - new reload test
* `configuration-as-code-with-templates-reduced.yml` - reduced YAML fixture

woeter69:
Just looking for some feedback/status of the review

Thanks!

PR #1498, opened 2026-03-15
Fix package name mismatch in test helper JAR

harshavardhanreddy06:
The email-ext-test-helper-0.1.jar contained a class with incorrect package structure (org 2 instead of org). This caused NoClassDefFoundError during test compilation. Recreated the jar with correct package path.


ArpanC6:
Hi @harshavardhanreddy06! Clean fix — the package name mismatch in the test helper JAR is resolved and all 17 CI checks are passing.
One suggestion: since this is a binary JAR committed to the repository, it would be helpful to document in the PR description how the JAR was recreated (source location, build command). This helps future maintainers regenerate it if needed.
Overall looks good!

slide:
I'm curious why this issue is not seen in normal testing, can you give additional information on the failure you saw that you feel necessitated this fix?

harshavardhanreddy06:
Hi @slide , thanks for the question!

I found this issue while working on the CI pipeline — the tests were failing on the Windows CI runner but passing fine on my local machine (Linux/macOS).

The problem was with the email-ext-test-helper-0.1.jar file. Inside a JAR, classes are stored in folders that must exactly match the package name. In this JAR, one of the folder names had a typo — it had org 2 (with a space) instead of just org. So the JVM couldn't find the class when it looked for it, and threw a NoClassDefFoundError.

The reason it doesn't show up in normal testing is that this JAR isn't loaded when you compile the code — it only gets loaded at runtime when a specific test runs a Groovy script. So if you happen to run tests in an environment that's a bit more forgiving about that space in the path (like Linux), it might quietly work or skip the error. But on Windows, the JVM is stricter about it and fails clearly.

The fix was simple: I recreated the JAR with the correct folder structure so the class can be found reliably on all platforms.

Let me know if you'd like me to add details on how the JAR was recreated to the PR description!

slide:
My main development environment is Windows, and I don't see any issue, can you share more information on how to reproduce the problem?

PR #1499, opened 2026-03-15
Fix inline CID attachments by using multipart/related to avoid duplicate attachment previews

Philodoescode:
This change improves how inline attachments are sent so clients like Microsoft Teams render CID images correctly without also showing them as normal attachment previews. #1438 #1465 

Changes:
- Refactored email body construction in `ExtendedEmailPublisher` to build a MIME tree with:
  - root: `multipart/mixed`
  - nested body+inline block: `multipart/related`
- Moved body creation into `createTextContentPart(...)` to make it reusable in the new MIME structure.
- Added inline attachments to the `multipart/related` section instead of the root mixed section.
- Kept regular attachments and build-log attachments on the root `multipart/mixed`.
- Updated `AttachmentUtils` to only set `Content-ID` for inline attachments (not regular attachments).
- Preserved support for project-level and trigger-level inline attachment patterns.

note: I apologize that the branch merging from is the main branch and not the feature branch from my forked repo.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
+++ b/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
@@ -163,7 +163,9 @@ public class AttachmentUtils implements Serializable {
                     try {
                         attachmentPart.setDataHandler(new DataHandler(fileDataSource));
                         attachmentPart.setFileName(MimeUtility.encodeText(file.getName()));
-                        attachmentPart.setContentID("<%s>".formatted(file.getName()));
+                        if (isInline) {
+                            attachmentPart.setContentID("<%s>".formatted(file.getName()));
+                        }
                         attachmentPart.setDisposition(isInline ? Part.INLINE : Part.ATTACHMENT);
                         attachments.add(attachmentPart);
                         totalAttachmentSize += file.length();
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -995,39 +995,51 @@ public class ExtendedEmailPublisher extends Notifier {
         msg.setSentDate(new Date());
         setSubject(context, msg, charset);
 
-        Multipart multipart = addContent(context, charset);
+        MimeMultipart rootMultipart = new MimeMultipart("mixed");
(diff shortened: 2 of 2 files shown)


ArpanC6:
Hi @Philodoescode! The CI failure is in ExtendedEmailPublisherTest.testPlainTextAndHtml — this test is likely failing because the MIME structure change (adding multipart/related wrapper) broke an existing assertion about the email structure.
You'll need to update the test to expect the new MIME tree:

Root: multipart/mixed
Nested: multipart/related (body + inline attachments)
Regular attachments stay on root multipart/mixed

Also note: the PR is merging from your main branch instead of a feature branch — this is not ideal practice. Consider creating a dedicated feature branch for cleaner history.
The Windows CI passed with 2 tests, so the core logic seems correct. Hope this helps!

PawanTheGod:
Hi 👋

I took a closer look at the failing test (`testPlainTextAndHtml`), and it seems like the failure is due to the test still expecting the old MIME structure.

Since this change introduces a nested `multipart/related` structure for inline content, the assertions in the test likely need to be updated to reflect this new layout.

I’d be happy to help update the test accordingly if that sounds good.

Thanks!

slide:
Is this a follow-on to #1465?

Philodoescode:
> Is this a follow-on to #1465?

Yes, it is

slide:
Thanks for the context — as a follow-on to #1465 this makes more sense. The `Content-ID` guard and the `setDisposition` fix for regular attachments are genuine improvements that #1465 doesn't address.

However the `createMail` restructuring in this PR significantly overlaps with what #1465 already does — both PRs rework the same `mixed`/`related` multipart structure in the same region. Once #1465 merges, this PR will need a rebase to strip out the overlapping structural changes, leaving only:

1. The `Content-ID` guard in `AttachmentUtils.java` — regular attachments shouldn't get `Content-ID` headers
2. The `setDisposition` fix ensuring `Part.ATTACHMENT` is always set for regular attachments

After that rebase it should be a small, clean follow-on. As it stands it can't merge without conflict resolution against #1465.

### No tests for the new behaviour

The `Content-ID` guard and disposition fix are the core contributions here — a test asserting that regular attachments have `Part.ATTACHMENT` disposition and no `Content-ID` header would lock in this behaviour and be a good addition alongside the rebase.

PR #1502, opened 2026-03-16
Prioritize project recipientList over trigger recipientList (Resolves JENKINS-71828)

harshavardhanreddy06:
## Summary

Fixes [JENKINS-71828](https://issues.jenkins.io/browse/JENKINS-71828) where specifying a "Project Recipient List" at the project level failed to override the default recipient list set on the trigger.

## Changes

- Modified [createMail](cci:1://file:///Users/harshareddy/Desktop/email-ext-plugin/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java:970:4-1148:5) in [ExtendedEmailPublisher.java](cci:7://file:///Users/harshareddy/Desktop/email-ext-plugin/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java:0:0-0:0) to evaluate an `effectiveRecipientList` before resolving the final address list.
- We now prioritize `this.recipientList` (the project recipient list config) using `StringUtils.isNotBlank()`. If the project recipient list is blank/unspecified, it gracefully falls back to `context.getTrigger().getEmail().getRecipientList()`, preserving existing configurations and maintaining backwards compatibility.

## Testing Done

- Validated code compiles successfully via `mvn clean compile`.
- Checked backward compatibility using ternary logic for graceful configuration fallback.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -1057,6 +1057,7 @@ public class ExtendedEmailPublisher extends Notifier {
             }
 
             String effectiveRecipientList = StringUtils.isNotBlank(this.recipientList)
+                    && !"$DEFAULT_RECIPIENTS".equals(this.recipientList)
                     ? this.recipientList
                     : context.getTrigger().getEmail().getRecipientList();
             descriptor.debug(context.getListener().getLogger(), "Adding recipients from recipient list");
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -444,7 +444,9 @@ public class ExtendedEmailPublisher extends Notifier {
         for (EmailTrigger trigger : getConfiguredTriggers()) {
             if (trigger.isPreBuild() == forPreBuild && trigger.trigger(build, listener)) {
                 EmailTriggerDescriptor triggerDescriptor = trigger.getDescriptor();
-                String tName = triggerDescriptor != null ? triggerDescriptor.getDisplayName() : trigger.getClass().getSimpleName();
+                String tName = triggerDescriptor != null
+                        ? triggerDescriptor.getDisplayName()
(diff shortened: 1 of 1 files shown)


ArpanC6:
Hi @harshavardhanreddy06! I noticed the CI failure in `TriggerNameContentTest.testTriggerName`. 

This test might be failing because the recipient list logic change is affecting how trigger names are resolved. You may want to check if the `effectiveRecipientList` change in `createMail()` is impacting the trigger context in any way.

You can run the specific test locally with:
mvn test -Dtest=TriggerNameContentTest#testTriggerName

Hope this helps!

slide:
I'm not sure this is the direction I want to go with this. I will need to review the bug and your fix in more depth.

harshavardhanreddy06:
Hi @slide, no worries at all — Happy to adjust the approach based on your feedback once you've had a chance to review. Let me know if you have any questions or want me to try a different direction.

slide:
Thanks for the null safety improvements — the fallback to `trigger.getClass().getSimpleName()` and the `getTriggerReplaceList()` guard are both sensible defensive fixes.

A couple of things to address:

### The `$DEFAULT_RECIPIENTS` recipient logic appears to be from a different PR

Lines 108–122 add recipient list override logic that also appears in #1576. This looks like it may have ended up here because both PRs were developed on the same branch rather than separate ones. This change is out of scope for a null-safety PR and inherits all the open concerns from #1576 — the string check is fragile and there are no tests for this behaviour. Please rebase this PR onto a fresh branch from `main` to remove the unrelated changes, and keep that logic in #1576 where it can be reviewed in context.

### Null descriptor should be logged, not silently swallowed

`getDescriptor()` returning `null` shouldn't happen in normal Jenkins operation — it would indicate a seriously broken plugin installation. Silently falling back to `"UTF-8"` or `""` may hide real problems from operators. A `LOGGER.warning()` when `descriptor()` returns `null` would make the anomaly visible.

### `descriptor()` vs `getDescriptor()` — please explain the distinction

The diff switches from `getDescriptor()` to a new `descriptor()` call without explanation. If `descriptor()` is a new method that can legitimately return `null` (unlike the Jenkins API `getDescriptor()`), that distinction should be documented with a Javadoc comment.

PR #1513, opened 2026-03-18
Validation for SmtpServer, hostname and/or IP

7assan2yman:
- Adds syntax validation for SMTP server hostnames in MailAccount.isSmtpServerValid(), allowing empty values for localhost.
- Checks non‑blank hostnames against RFC‑compliant patterns for hostnames and IP addresses.
### Testing done
Ran mvn spotless:apply
Ran mvn verify -DskipTests

This my first PR. Sorry If I made any mistakes
### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -93,19 +93,20 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
     }
 
     private boolean isSyntacticallyValid(String host) {
-        // Hostname pattern (RFC 1123): letters, digits, hyphens, dots
-        Pattern HOSTNAME_PATTERN = Pattern.compile(
+        // Check if it looks like an IPv4 address (four dot‑separated numbers)
+        java.util.regex.Pattern ipv4Structure = java.util.regex.Pattern.compile("^(\\d{1,3}\\.){3}\\d{1,3}$");
+        if (ipv4Structure.matcher(host).matches()) {
+            // Strict IPv4 validation
+            java.util.regex.Pattern ipv4Valid = java.util.regex.Pattern.compile(
+                    "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
+            );
+            return ipv4Valid.matcher(host).matches();
+        }
+        // Otherwise treat as hostname (RFC 1123)
+        java.util.regex.Pattern hostnameValid = java.util.regex.Pattern.compile(
                 "^(?![0-9]+$)(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*$"
         );
-        // IPv4 pattern
-        Pattern IPV4_PATTERN = Pattern.compile(
-                "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
-        );
-        return HOSTNAME_PATTERN.matcher(host).matches()
-                || IPV4_PATTERN.matcher(host).matches();
-        // IPv6 is not covered by fallback – but if the host was a valid IPv6 literal,
-        // InetAddress.getByName would have succeeded in a real environment.
-        // For CI tests that never use IPv6, this is fine.
+        return hostnameValid.matcher(host).matches() && host.length() <= 255;
     }
 
     public boolean isSmtpAuthValid() {
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -86,14 +86,12 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
     }
 
     private boolean isHostnameValid(String host) {
-        String hostnamePattern =
-                "^(?![0-9]+$)(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*$";
+        String hostnamePattern = "^(?![0-9]+$)(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*$";
         return host.matches(hostnamePattern);
     }
 
     private boolean isIpAddressValid(String host) {
-        String ipPattern =
-                "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$";
+        String ipPattern = "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$";
         return host.matches(ipPattern);
     }
 
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -77,10 +77,24 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
     }
 
     public boolean isSmtpServerValid() {
-        return true;
-        // Note: having no SMTP server is fine, it means localhost.
-        // More control could be implemented here when not null though,
-        // like checking the value looks like an FQDN or IP address.
+        // Empty means "localhost", which is always valid.
+        if (StringUtils.isBlank(smtpHost)) {
+            return true;
+        }
+        // Validate hostname or IP address syntax.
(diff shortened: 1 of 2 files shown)


ArpanC6:
Hi @7assan2yman! The implementation looks good overall — validating SMTP hostname syntax before attempting a connection is solid defensive programming. A few observations after reviewing the code:
Strengths:

Clean separation into isHostnameValid() and isIpAddressValid() — easy to read and maintain
Correctly returns true for blank host (localhost fallback is valid behavior)
Regex patterns look correct for standard IPv4 and hostnames

Suggestions:
1. IPv6 support missing:
isIpAddressValid() only handles IPv4. Some enterprise Jenkins installations use IPv6 SMTP servers (e.g., ::1 or 2001:db8::1). The current regex will reject these as invalid.
A simpler alternative that handles IPv4, IPv6, and hostnames in one call — already used elsewhere in this codebase (PR #1471):
try {
    InetAddress.getByName(smtpHost);
    return true;
} catch (UnknownHostException e) {
    return false;
}
2. Missing unit tests:
CI report shows lines 94-95 are not covered. Consider adding tests for:

Valid hostname: smtp.gmail.com
Valid IP: 192.168.1.1
Invalid hostname: invalid..host
Invalid IP: 999.999.999.999
Blank/null (already returns true)

3. Minor — partial branch coverage on line 85:
CI warns about 3 missing branches on return isHostnameValid(smtpHost) || isIpAddressValid(smtpHost). Unit tests above will fix this automatically.
All CI checks are passing otherwise. Great first PR!

7assan2yman:
> Hi @7assan2yman! The implementation looks good overall — validating SMTP hostname syntax before attempting a connection is solid defensive programming. A few observations after reviewing the code: Strengths:
> 
> Clean separation into isHostnameValid() and isIpAddressValid() — easy to read and maintain Correctly returns true for blank host (localhost fallback is valid behavior) Regex patterns look correct for standard IPv4 and hostnames
> 
> Suggestions:
> 
>     1. IPv6 support missing:
>        isIpAddressValid() only handles IPv4. Some enterprise Jenkins installations use IPv6 SMTP servers (e.g., ::1 or 2001:db8::1). The current regex will reject these as invalid.
>        A simpler alternative that handles IPv4, IPv6, and hostnames in one call — already used elsewhere in this codebase (PR [Improve SMTP host and port validation with descriptor checks and unit… #1471](https://github.com/jenkinsci/email-ext-plugin/pull/1471)):
>        try {
>        InetAddress.getByName(smtpHost);
>        return true;
>        } catch (UnknownHostException e) {
>        return false;
>        }
> 
>     2. Missing unit tests:
>        CI report shows lines 94-95 are not covered. Consider adding tests for:
> 
> 
> Valid hostname: smtp.gmail.com Valid IP: 192.168.1.1 Invalid hostname: invalid..host Invalid IP: 999.999.999.999 Blank/null (already returns true)
> 
>     3. Minor — partial branch coverage on line 85:
>        CI warns about 3 missing branches on return isHostnameValid(smtpHost) || isIpAddressValid(smtpHost). Unit tests above will fix this automatically.
>        All CI checks are passing otherwise. Great first PR!

You're absolutely right. InetAddress.getByName(smtpHost) is a simpler approach. I will write the unit tests and push changes shortly.

7assan2yman:
Somehow using InetAddress.getByName breaks the CI. There's another option which is using InetAddressValidator. However, this requires adding Apache Commons dependency. I'm considering falling back to using regex validation (without ipv6 support). Any guidance or help is appreciated.

ArpanC6:
@7assan2yman! The CI failure with InetAddress.getByName() 
is likely due to UnknownHostException being thrown for valid 
hostnames in the offline CI environment.

The root cause: CI has no internet access, so InetAddress 
cannot perform DNS lookup  even for valid hostnames like 
smtp.gmail.com. This causes UnknownHostException even when 
the hostname itself is correct.

A simple fix  wrap it with a try catch that also handles 
the offline case:

try {
    InetAddress.getByName(smtpHost);
    return true;
} catch (UnknownHostException e) {
    // Fall back to regex for offline CI environments
    return HOSTNAME_PATTERN.matcher(smtpHost).matches() 
        || IP_PATTERN.matcher(smtpHost).matches();
}

This way:
1)  Real environment: InetAddress handles IPv4, IPv6, and 
  hostnames in one call
2)  CI (offline): regex fallback validates standard hostnames 
  and IPv4

IPv6 support is preserved when network is available and 
no extra dependency needed..

7assan2yman:
Thanks for the suggestion. I will try that.

7assan2yman:
@ArpanC6 it worked. Thanks a lot for the help.

slide (changes requested):

slide:
Good that the long-standing TODO is finally being addressed, and the two-phase approach is sensible. A few issues worth addressing though:

### Live DNS lookup is problematic 

`InetAddress.getByName(smtpHost)` makes a real network call. In air-gapped environments or when the SMTP server is on a private network not yet in DNS, this will return `false` and fall through to `isSyntacticallyValid()` — which may also return `false` for a perfectly valid hostname, incorrectly blocking configuration. Syntax-only validation should be the primary check, with DNS lookup as an optional enhancement at most.

### IPv6 validation is missing from `isSyntacticallyValid()`

The tests include IPv6 cases (`::1`, `2001:db8::1`) but `isSyntacticallyValid()` has no IPv6 handling. Both would fail the IPv4 check and the hostname regex, returning `false`. The IPv6 tests only pass because `InetAddress.getByName()` resolves them successfully first — so in a DNS-unavailable scenario, valid IPv6 addresses would incorrectly fail validation. An IPv6 regex path needs to be added to `isSyntacticallyValid()`.

### `Pattern.compile` inside instance methods

All three patterns are compiled fresh on every call. They should be `private static final` fields:

```java
private static final Pattern IPV4_STRUCTURE = Pattern.compile("^(\\d{1,3}\\.){3}\\d{1,3}$");
private static final Pattern IPV4_VALID = Pattern.compile(
    "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$");
private static final Pattern HOSTNAME_VALID = Pattern.compile(
    "^(?![0-9]+$)(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*$");
```

### IPv6 tests rely on live DNS in a `@WithoutJenkins` test

```java
account.setSmtpHost("::1");
assertTrue(account.isSmtpServerValid());
```

This passes because `InetAddress.getByName("::1")` resolves on most systems, but it's a network call in what should be an isolated unit test. Once IPv6 handling is added to `isSyntacticallyValid()` and DNS lookup is made secondary, these tests will work without any network dependency.

7assan2yman:
> Good that the long-standing TODO is finally being addressed, and the two-phase approach is sensible. A few issues worth addressing though:
> ### Live DNS lookup is problematic
> 
> `InetAddress.getByName(smtpHost)` makes a real network call. In air-gapped environments or when the SMTP server is on a private network not yet in DNS, this will return `false` and fall through to `isSyntacticallyValid()` — which may also return `false` for a perfectly valid hostname, incorrectly blocking configuration. Syntax-only validation should be the primary check, with DNS lookup as an optional enhancement at most.
> ### IPv6 validation is missing from `isSyntacticallyValid()`
> 
> The tests include IPv6 cases (`::1`, `2001:db8::1`) but `isSyntacticallyValid()` has no IPv6 handling. Both would fail the IPv4 check and the hostname regex, returning `false`. The IPv6 tests only pass because `InetAddress.getByName()` resolves them successfully first — so in a DNS-unavailable scenario, valid IPv6 addresses would incorrectly fail validation. An IPv6 regex path needs to be added to `isSyntacticallyValid()`.
> ### `Pattern.compile` inside instance methods
> 
> All three patterns are compiled fresh on every call. They should be `private static final` fields:
> 
> ```java
> private static final Pattern IPV4_STRUCTURE = Pattern.compile("^(\\d{1,3}\\.){3}\\d{1,3}$");
> private static final Pattern IPV4_VALID = Pattern.compile(
>     "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$");
> private static final Pattern HOSTNAME_VALID = Pattern.compile(
>     "^(?![0-9]+$)(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*$");
> ```
> 
> ### IPv6 tests rely on live DNS in a `@WithoutJenkins` test
> 
> ```java
> account.setSmtpHost("::1");
> assertTrue(account.isSmtpServerValid());
> ```
> 
> This passes because `InetAddress.getByName("::1")` resolves on most systems, but it's a network call in what should be an isolated unit test. Once IPv6 handling is added to `isSyntacticallyValid()` and DNS lookup is made secondary, these tests will work without any network dependency.

Hello
I think there's no need for `isSyntacticallyValid` to  validate IP addresses at all (ipv4 and ipv6). According to [this](https://docs.oracle.com/javame/8.0/api/gcf/api/javax/microedition/io/NetworkUtilities.html) "If a literal IP address is supplied, only the validity of the address format is checked". As a result, we should only validate hostnames (the only thing that trigger a DNS check if passed to InetAddress.getByName).

I'm happy to make any other needed edits.

slide (approved):

PR #1537, opened 2026-03-27
fix:handle edge case for empty recipients to prevent crash

Rashmi1613:
Previously, if `msg.getAllRecipients()` returned null or an empty array,
the code attempted to access `allRecipients[0]`, which could lead to
NullPointerException or ArrayIndexOutOfBoundsException.

This change adds a safety check to prevent crashes and logs a clear message instead.

---

### Changes Made

- Added null and empty check for `allRecipients`
- Prevented unsafe access to `allRecipients[0]`
- Added log message for better debugging

---

### Testing

- Verified that when recipients are empty, email is not sent
- Confirmed no crash occurs in this scenario


ArpanC6:
I went through the diff and noticed a few things

1. Compile error missing parentheses
This is probably why CI is failing. **getListener** and **getLogger** are methods not fields so they need parentheses:

**context.getListener().getLogger().println(...)**

2) Logging approach
**getLogger().println()** writes to the build console, not the plugin logs. The rest of the codebase seems to be using **java.util.logging.Logger** so it might be better to stay consistent with that.

3) Check is happening too early
Since **executePresendScript()** can modify the recipient list at runtime, checking **getAllRecipients()** before it runs might not give the final result. It would be safer to move the null/empty check after **executePresendScript().**

4) Log message formatting
Small thing there’s a missing space after the comma in **"No recipients** **found,email** couldn't be sent."

5) Stray debug comment
Looks like there’s a debug comment left in the diff probably worth removing before pushing.

Overall the fix definitely makes sense.
Once the compile issue and the presend script ordering are sorted, this should be in good shape..

Rashmi1613:
@ArpanC6  Thanks for the feedback! I have moved the recipient check after executePresendScript(). I have kept the current logging approach for consistency ,but happy to update if preferred.

slide:
Thanks for working on this — guarding against a null/empty `allRecipients` before accessing `allRecipients[0]` is a real crash risk worth fixing.

That said, there are a few things I think need to be addressed before this is ready to merge:

### Logic error in the empty-recipient check

The null/empty check on `allRecipients` still fires before the result of the emergency reroute is reflected. `allRecipients` is fetched before the reroute block runs, so even when an emergency reroute is configured and would supply a valid recipient, the early return kicks in and aborts. The fix should either re-fetch `allRecipients` after the reroute block, or gate the null check on the reroute not being configured.

### `logDuplicateRecipients()` appears to be out of scope

This method wasn't mentioned in the PR description and looks like it may have come from a separate branch. It also has some issues:
- The BCC `removeIf` lambda returns `false` when a duplicate CC address is found, which means it logs the duplicate but doesn't actually remove it.
- The duplicate detection logic runs a second time in `createMail` via the `emailLocations` map — so the same work is done twice, inconsistently.

I'd suggest splitting this out into its own PR so it can get proper review.

### Missing test coverage for the new guard

It would be good to have a test that verifies `sendMail` returns `false` and logs the expected message when `getAllRecipients()` returns `null` or empty.

### Commit history

There are 10 commits here with several generic messages like `changes made` and `fixed formatting`. Please squash these down before merging — a single focused commit for the null-check fix would be ideal.

PR #1538, opened 2026-03-27
fix:improve logging for smtp session creation failure

Rashmi1613:
The log message before did not provide much context on why the bug was appearing.
Now it includes the smtp server information when smtp session creation fails.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -570,8 +570,10 @@ public class ExtendedEmailPublisher extends Notifier {
             }
             Session session = getDescriptor().createSession(mailAccount, context);
             if (session == null) {
-                String host=mailAccount.getSmtpHost();
-                context.getListener().getLogger().println("Could not create session"+(host!=null ? host : "unknown") );
+                String host = mailAccount.getSmtpHost();
+                context.getListener()
+                        .getLogger()
+                        .println("Could not create session" + (host != null ? host : "unknown"));
                 return false;
             }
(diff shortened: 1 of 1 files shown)


ArpanC6:
Thanks for improving the log message adding SMTP host context will help administrators diagnose session failures faster.
Two issues I noticed - 

1) Missing space/separator - **"Could not create session"** + **mailAccount.getSmtpHost()** will produce output like **"Could not create sessionsmtp.office365.com"** with no space between the message and the host. Consider using **"Could not create session for SMTP host: "** + **mailAccount.getSmtpHost()** for clarity.

2) Missing null check -  If **getSmtpHost()** returns null, the log will show **"Could not create sessionnull".** Consider adding a null guard: **(mailAccount.getSmtpHost() != null ? mailAccount.getSmtpHost() : "unknown").**

Rashmi1613:
@ArpanC6 Thanks a lot for the feedback! I've updated the log message to include proper formatting and added the null check too

slide:
Thanks for keeping this focused — adding the SMTP host to the session failure message is a useful improvement for anyone trying to debug a misconfigured mail account.

A couple of things still need attention before this is ready to merge:

### Missing separator in the log message

The current log line:
```java
"Could not create session" + (host != null ? host : "unknown")
```
will produce output like `Could not create sessionsmtp.example.com` with no space or separator. ArpanC6 flagged this in their earlier comment but the fix only added the null guard. Consider changing it to:
```java
"Could not create session for SMTP host: " + (host != null ? host : "unknown")
```

### No test coverage

The CI check is flagging lines 573–576 as uncovered. A test that mocks `createSession` to return `null` and asserts the expected log output and `false` return value would be straightforward to add and would prevent future regression.

PR #1539, opened 2026-03-27
test: Add unit tests for RequesterRecipientProvider

Vinamra-tech:
RequesterRecipientProvider had no test coverage despite containing
non-trivial upstream cause traversal logic. Add four tests covering:
- Direct user trigger with no upstream cause receives email
- No email sent when no user and no upstream cause exists
- Original upstream user receives email when traversing upstream chain
- No crash when upstream project linkage is broken (graceful degradation)

### Testing done
All four tests pass locally via:
`mvn test -Dtest=RequesterRecipientProviderTest`

Output:
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0

### Submitter checklist

### DIFF
--- a/src/test/java/hudson/plugins/emailext/plugins/recipients/RequesterRecipientProviderTest.java
+++ b/src/test/java/hudson/plugins/emailext/plugins/recipients/RequesterRecipientProviderTest.java
@@ -1,0 +1,118 @@
+package hudson.plugins.emailext.plugins.recipients;
+
+import hudson.model.Cause;
+import hudson.model.FreeStyleBuild;
+import hudson.model.FreeStyleProject;
+import hudson.model.Job;
+import hudson.model.Result;
+import hudson.model.User;
+import hudson.plugins.emailext.ExtendedEmailPublisherDescriptor;
+import hudson.tasks.Mailer;
+import jenkins.model.Jenkins;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.mockito.MockedStatic;
+import org.mockito.Mockito;
+
+class RequesterRecipientProviderTest {
+
+    private MockedStatic<Jenkins> mockedJenkins;
+    private MockedStatic<Mailer> mockedMailer;
+    private Jenkins jenkins;
+
+    @BeforeEach
+    void before() {
+        jenkins = Mockito.mock(Jenkins.class);
+        Mockito.when(jenkins.isUseSecurity()).thenReturn(false);
+        final ExtendedEmailPublisherDescriptor descriptor =
+                Mockito.mock(ExtendedEmailPublisherDescriptor.class);
+        descriptor.setDebugMode(true);
+        Mockito.when(descriptor.getExcludedCommitters()).thenReturn("");
+        Mockito.when(jenkins.getDescriptorByType(ExtendedEmailPublisherDescriptor.class))
+                .thenReturn(descriptor);
+        mockedJenkins = Mockito.mockStatic(Jenkins.class);
+        mockedJenkins.when(Jenkins::get).thenReturn(jenkins);
+
+        final Mailer.DescriptorImpl mailerDescriptor = Mockito.mock(Mailer.DescriptorImpl.class);
+        Mockito.when(mailerDescriptor.getDefaultSuffix()).thenReturn("DOMAIN");
+        mockedMailer = Mockito.mockStatic(Mailer.class);
+        mockedMailer.when(Mailer::descriptor).thenReturn(mailerDescriptor);
+    }
+
+    @AfterEach
+    void after() {
+        mockedMailer.close();
+        mockedJenkins.close();
+    }
+
+    @Test
+    @DisplayName("User who directly triggered the build receives email when no upstream cause exists")
+    void testDirectUserTriggerReceivesEmail() throws Exception {
+        try (MockedStatic<User> mockedUser = Mockito.mockStatic(User.class)) {
+            final FreeStyleProject project = Mockito.mock(FreeStyleProject.class);
+            final FreeStyleBuild build = Mockito.spy(new FreeStyleBuild(project));
+            Mockito.doReturn(Result.FAILURE).when(build).getResult();
+            Mockito.doReturn(null).when(build).getCause(Cause.UpstreamCause.class);
+            MockUtilities.addRequestor(mockedUser, build, "A");
+            TestUtilities.checkRecipients(build, new RequesterRecipientProvider(), "A");
+        }
+    }
+
+    @Test
+    @DisplayName("No email sent when build was not triggered by a user and has no upstream cause")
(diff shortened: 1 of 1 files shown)


ArpanC6:
### 1) Windows CI failure
The failure on windows 21 (only 8 tests passing vs ~370 on Linux) usually points to the test not running inside a proper Jenkins test harness. Calling new FreeStyleBuild(project) directly can cause exactly this kind of issue especially on Windows where Jenkins isn’t fully initialized. It’d be safer to align this with how other tests in the same package are set up most likely using JenkinsRule or @JenkinsTest.

### 2) Missing edge case upstream build number not found
You’ve already covered the case where the upstream project is null in testBrokenUpstreamProjectLinkage(), which is great. But there’s still a small gap when the project exists but getBuildByNumber() returns null. That’s basically the other half of the broken linkage scenario so worth adding something like - 

Mockito.when(upstreamProject.getBuildByNumber(1)).thenReturn(null);
TestUtilities.checkRecipients(build, new RequesterRecipientProvider());

### 3) Minor debug mode on mock
descriptor.setDebugMode(true) doesn’t really do anything here since ExtendedEmailPublisherDescriptor is mocked it won’t retain state unless explicitly stubbed. Not a functional issue, just a bit misleading. You can either stub it properly or just remove the line to keep things clean.

Vinamra-tech:
> ### 1) Windows CI failure
> The failure on windows 21 (only 8 tests passing vs ~370 on Linux) usually points to the test not running inside a proper Jenkins test harness. Calling new FreeStyleBuild(project) directly can cause exactly this kind of issue especially on Windows where Jenkins isn’t fully initialized. It’d be safer to align this with how other tests in the same package are set up most likely using JenkinsRule or @jenkinstest.
> 
> ### 2) Missing edge case upstream build number not found
> You’ve already covered the case where the upstream project is null in testBrokenUpstreamProjectLinkage(), which is great. But there’s still a small gap when the project exists but getBuildByNumber() returns null. That’s basically the other half of the broken linkage scenario so worth adding something like -
> 
> Mockito.when(upstreamProject.getBuildByNumber(1)).thenReturn(null); TestUtilities.checkRecipients(build, new RequesterRecipientProvider());
> 
> ### 3) Minor debug mode on mock
> descriptor.setDebugMode(true) doesn’t really do anything here since ExtendedEmailPublisherDescriptor is mocked it won’t retain state unless explicitly stubbed. Not a functional issue, just a bit misleading. You can either stub it properly or just remove the line to keep things clean.

Thanks for detailed review , it's the second time you've made some genuine and interesting comments on my pr and that too back to back!!

1. Sure thing , that new FreeStyleBuild(project) triggered unitialised file system resulting in a crash which Linux forgives.To keep these as fast unit tests and avoid rewriting the whole suite to use JenkinsRule, I updated the setup to fully mock FreeStyleBuild (Mockito.mock(FreeStyleBuild.class)). This bypasses the constructor and isolates the tests entirely.

2. That' an excellent point!!!  I've added a 5th test (testUpstreamProjectExistsButBuildIsNotFound) to cover the scenario where the upstream project is found, but getBuildByNumber(1) returns null.

3. Sure, I just added the line  to be safe but actually testing it proves it's of no use.

Once again, Thank You for your time

ArpanC6:
> > ### 1) Windows CI failure
> > The failure on windows 21 (only 8 tests passing vs ~370 on Linux) usually points to the test not running inside a proper Jenkins test harness. Calling new FreeStyleBuild(project) directly can cause exactly this kind of issue especially on Windows where Jenkins isn’t fully initialized. It’d be safer to align this with how other tests in the same package are set up most likely using JenkinsRule or @jenkinstest.
> > ### 2) Missing edge case upstream build number not found
> > You’ve already covered the case where the upstream project is null in testBrokenUpstreamProjectLinkage(), which is great. But there’s still a small gap when the project exists but getBuildByNumber() returns null. That’s basically the other half of the broken linkage scenario so worth adding something like -
> > Mockito.when(upstreamProject.getBuildByNumber(1)).thenReturn(null); TestUtilities.checkRecipients(build, new RequesterRecipientProvider());
> > ### 3) Minor debug mode on mock
> > descriptor.setDebugMode(true) doesn’t really do anything here since ExtendedEmailPublisherDescriptor is mocked it won’t retain state unless explicitly stubbed. Not a functional issue, just a bit misleading. You can either stub it properly or just remove the line to keep things clean.
> 
> Thanks for detailed review , it's the second time you've made some genuine and interesting comments on my pr and that too back to back!!
> 
> 1. Sure thing , that new FreeStyleBuild(project) triggered unitialised file system resulting in a crash which Linux forgives.To keep these as fast unit tests and avoid rewriting the whole suite to use JenkinsRule, I updated the setup to fully mock FreeStyleBuild (Mockito.mock(FreeStyleBuild.class)). This bypasses the constructor and isolates the tests entirely.
> 2. That' an excellent point!!!  I've added a 5th test (testUpstreamProjectExistsButBuildIsNotFound) to cover the scenario where the upstream project is found, but getBuildByNumber(1) returns null.
> 3. Sure, I just added the line  to be safe but actually testing it proves it's of no use.
> 
> Once again, Thank You for your time


### Point 1 - You’re absolutely right about mocking FreeStyleBuild instead of using new FreeStyleBuild(project). The key issue with using the constructor directly is that it triggers Jenkins' internal chain which tries to interact with the filesystem. This is the main culprit behind the Windows CI crash. By mocking it fully you avoid that and keep things isolated which makes the test purely a unit test without any external dependencies. This approach definitely feels cleaner and safer.

### Point 2 - I’m really glad that resonated with you. The test for testUpstreamProjectExistsButBuildsNotFound is an important one. In real production scenarios it's totally possible for build records to be missing (due to manual deletions retention policies etc) even if the project itself is still there. So having that case explicitly covered in the tests helps make sure the code handles it gracefully when it comes up in the wild.

### Point 3 - I agree removing the silent state change in the mock makes the code clearer. If you ever need to test debug logging behavior it’s much better to stub it properly like with Mockito.when(descriptor.isDebugMode()).thenReturn(true). Directly setting state on a mock can be really confusing for anyone reading the code later and it doesn’t actually have any effect on the test. So the cleaner approach is definitely the way to go.

slide:
Good set of tests — the scenarios are well-chosen and cover the main logical paths through the upstream cause traversal logic nicely. The 5th test (`testUpstreamProjectExistsButBuildIsNotFound`) was a good addition.

A couple of minor things worth tidying up:

### Unused local variables

`final FreeStyleProject project = Mockito.mock(FreeStyleProject.class)` is declared in several tests but never actually used — the tests operate entirely on the mocked `build`. These can be removed.

### `testNoEmailWhenNoUserAndNoUpstreamCause` may not be testing what it claims

This test stubs `getCause(Cause.UserIdCause.class)` to return `null`, but since `MockUtilities.addRequestor` is never called, no user is set up to begin with. It's worth checking whether the provider actually reaches the `UserIdCause` code path when `UpstreamCause` is already `null` — if it doesn't, the stub is redundant and the test name is slightly misleading. Renaming to something like `testNoEmailWhenBuildHasNoCause` might be more accurate, or alternatively add a separate test that specifically exercises the `UserIdCause` null path.

PR #1545, opened 2026-03-29
Prevent Integer Overflow and NPE in FailedTestsContent

Vinamra-tech:
This PR addresses an issue in `FailedTestsContent.java` where a `NullPointerException` could occur when generating email content for failed tests. 

Specifically, if a failed test had a `null` value for its error details or stack trace, passing it directly to `StringEscapeUtils.escapeHtml4()` while `escapeHtml` was enabled would cause an exception. 

Changes made:
* Added explicit `null` checks for `rawStack` and `rawDetails` before attempting to escape them as HTML.
* Updated the `maxLength` parameter and its internal limits from `int` to `long` (`Long.MAX_VALUE`) to handle larger output sizes safely and prevent potential integer overflow.

### Testing done

* Executed the `FailedTestsContentTest` suite locally via Maven Surefire, with all 14 tests passing successfully.
* Added a new automated unit test (`testGetContent_withMessage_withStack_htmlEscaped_nullValues`) specifically to verify that `null` error details and stack traces are handled gracefully without throwing a `NullPointerException` when HTML escaping is enabled.
* Updated `testGetContent_whenContentLargerThanMaxLengthShouldTruncate` to test the new `long` boundary for `maxLength`.

WARNING : The @Parameter-annotated public field maxLength has been changed from int to long to prevent integer overflow. This constitutes a breaking change for any downstream consumers or scripts that specifically read this field back as an int.

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira *(Please add the Jira ID here if applicable)*

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.AbstractBuild;
@@ -42,6 +43,7 @@ public class FailedTestsContent extends DataBoundTokenMacro {
     public boolean onlyRegressions = false;
 
     @Parameter
+    @SuppressFBWarnings(value = "PA_PUBLIC_PRIMITIVE_ATTRIBUTE", justification = "TODO needs triage")
     public long maxLength = Long.MAX_VALUE;
 
     @Parameter
--- a/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/content/FailedTestsContent.java
@@ -5,7 +5,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.AbstractBuild;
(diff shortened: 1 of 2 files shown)


ArpanC6:
### Hi @Vinamra-tech good fix for the NPE! The null checks and the new test look solid. One issue I noticed in setMaxLength() the condition maxLength < Long.MAX_VALUE doesn't fully prevent overflow. If maxLength is between Long.MAX_VALUE / 1024 and Long.MAX_VALUE, then maxLength *= 1024 will still overflow. Consider changing the condition to maxLength < Long.MAX_VALUE / 1024 to guard against this edge case.

Vinamra-tech:
> ### Hi @Vinamra-tech good fix for the NPE! The null checks and the new test look solid. One issue I noticed in setMaxLength() the condition maxLength < Long.MAX_VALUE doesn't fully prevent overflow. If maxLength is between Long.MAX_VALUE / 1024 and Long.MAX_VALUE, then maxLength *= 1024 will still overflow. Consider changing the condition to maxLength < Long.MAX_VALUE / 1024 to guard against this edge case.

Sure thing!! I updated the code to be more robust against Overflow for which the pr was opened in the first place.

slide:
Two genuine bugs fixed here and the fixes are clean — the NPE guard and the overflow check are both handled correctly. Nearly there, just a few things worth noting:

### `setMaxLength()` semantics for the default value

The original code skipped the multiplication when `maxLength == Integer.MAX_VALUE`, treating it as "no limit". The new ternary handles the equivalent case correctly (keeping `Long.MAX_VALUE` when it would overflow), but it's non-obvious at a glance. A short comment explaining the intent would help future readers:

```java
// Multiply by 1024 to convert KB to bytes; cap at Long.MAX_VALUE to prevent overflow
maxLength = (maxLength <= Long.MAX_VALUE / 1024) ? maxLength * 1024 : Long.MAX_VALUE;
```

### `maxLength` type change is a public API change

Since `maxLength` is a `@Parameter`-annotated public field, changing it from `int` to `long` is technically a breaking change for anything that reads it back as `int`. It's worth calling this out explicitly in the PR description so maintainers can make an informed decision about compatibility.

### The new test could assert more explicitly

`testGetContent_withMessage_withStack_htmlEscaped_nullValues` verifies the output matches the expected string, which is good. It would be even clearer to also assert that no stack trace or error detail content appears in the output, since both were null — making the intent of the null-guard explicit rather than implicit.

Vinamra-tech:
@slide ,I've addressed all your concerns in points below:

I've added the suggested explanatory comment above the ternary operator to clarify the overflow prevention logic.

I've updated the main PR description to explicitly call out the int to long type change for the @Parameter maxLength field so maintainers are aware of the breaking change for downstream consumers.

I've updated testGetContent_withMessage_withStack_htmlEscaped_nullValues to explicitly assert that the generated HTML does not contain any "Error Message:" or "Stack Trace:" content when those values are null.

PR #1550, opened 2026-04-01
Extract SMTP property keys as named constants in ExtendedEmailPublisherDescriptor

ArpanC6:
### Summary

This change extracts hardcoded SMTP property key strings in 
ExtendedEmailPublisherDescriptor.java into private static final 
String constants at the class level.

### Why this change?

Earlier strings like "mail.smtp.host", "mail.smtp.auth", 
"mail.smtp.starttls.enable" etc. were directly hardcoded inside 
the createSession() method. There were also two local constants 
(SMTP_PORT_PROPERTY and SMTP_SOCKETFACTORY_PORT_PROPERTY) defined 
inside the method itself which limited their scope.

### What's changed?

- Introduced 12 private static final String constants at the 
  class level
- Moved the two existing local constants to the class level 
  for consistency
- Replaced all hardcoded SMTP property strings with these constants

### Behavioral changes?

None. This is purely a refactor/cleanup.

This is a clean resubmission of #1536. That PR accidentally 
included commits from other branches. This PR contains only 
the constants extraction change.


ArpanC6:
@slide Closing this in favor of #1550 which is a clean resubmission containing only the constants extraction change.

PR #1552, opened 2026-04-02
Warn when both SSL and TLS are enabled

Vaibhav34-2006:
This change adds a configuration warning when both SSL and TLS are enabled for a mail account.

Previously, `doCheckCredentialsId(...)` returned `FormValidation.ok()` when both `useSsl` and `useTls` were set to `true`.

Since enabling both options simultaneously is generally unintended and may result in ambiguous SMTP configuration, the method now returns `FormValidation.warning(...)` with an explanatory message:

`Both SSL and TLS are enabled. Usually only one should be selected.`

### Testing done

Updated the existing validation test in `MailAccountFIPSTest` to verify that `doCheckCredentialsId(...)` returns `FormValidation.Kind.WARNING` when both SSL and TLS are enabled together while valid credentials are configured.

Verified that:

* `useSsl=true` and `useTls=false` still returns `OK`

* `useSsl=false` and `useTls=true` still returns `OK`

* `useSsl=true` and `useTls=true` with credentials now returns `WARNING`

* The no-auth case still returns `OK`

* [x] Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch
* [x] Ensure the pull request title represents the desired changelog entry
* [x] Please describe what you did
* [ ] Link to relevant issues in GitHub or Jira
* [ ] Link to relevant pull requests, esp. upstream and downstream changes
* [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -126,7 +126,7 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
             }
             return result.includeEmptyValue()
                     .includeMatchingAs(
-                            item instanceof Queue.Task t ? Tasks.getAuthenticationOf(t) : ACL.SYSTEM,
+                            item instanceof Queue.Task t ? Tasks.getAuthenticationOf(t) : Jenkins.getAuthentication(),
                             item,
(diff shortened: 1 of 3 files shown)


Vaibhav34-2006:
Remove unrelated MailAccount changes and keep only SSL/TLS warning logic

PR #1553, opened 2026-04-02
fix: restore thread interrupt flag in renderTemplate() and split broad exception handling

ArpanC6:
### What was wrong

In the method EmailExtTemplateAction.renderTemplate(), there was a broad catch (Exception ex) block that was catching both InterruptedException and all other exceptions. However it didn’t handle the thread interruption correctly. Specifically when an InterruptedException is caught the JVM clears the thread’s interrupt flag automatically but if the code doesn't call Thread.currentThread().interrupt() the interrupt status gets lost. This can lead to a situation where the interruption is ignored, and other parts of the system (like upstream callers that check the thread's interrupted status) won't know that the thread was actually interrupted.

This is a common concurrency bug in Java that can affect long running systems like Jenkins. If a thread is interrupted (for instance, during a graceful shutdown or job cancellation) but the interrupt flag is silently cleared the thread might continue running instead of shutting down properly.

The same issue was present in doTemplateFileCheck() where IOException and InterruptedException were caught in a single block again without restoring the interrupt flag.

### What I changed

In EmailExtTemplateAction.java I split the catch (Exception ex) block into two separate blocks:

catch (InterruptedException ex): This block now restores the thread’s interrupt flag by calling Thread.currentThread().interrupt() then proceeds with rendering the error message.
catch (Exception ex): This block continues to handle all other exceptions as before with no changes in behavior for non-interrupt exceptions.

Similarly I split the catch (IOException | InterruptedException e) block in doTemplateFileCheck() into two blocks, following the same pattern of restoring the interrupt flag for InterruptedException.

### Why this matters

Jenkins handles many concurrent builds and if a build thread is interrupted (like during a graceful shutdown or job abort) but the interrupt flag is swallowed quietly the thread might keep running instead of stopping as expected. By restoring the interrupt flag I ensure that interruptions propagate properly and the thread responds to shutdown signals correctly.

### Test added

I also added a new test class EmailExtTemplateActionTest that:

Sets the thread’s interrupt flag before calling renderTemplate().
Calls renderTemplate() with a non existent template file.
Verifies that the interrupt flag is restored after the method finishes.

Result: Tests run: 1, Failures: 0, Errors: 0.


ArpanC6:
Hlw @slide just wanted to give you a quick heads up on this PR and the reasoning behind it.

While I was going through the code I noticed that in renderTemplate() (in EmailExtTemplateAction) there was a broad catch (Exception ex) block that was silently swallowing InterruptedException and not restoring the interrupt flag. In Java when an InterruptedException is caught the JVM clears the interrupt flag automatically. If the code doesn’t call Thread.currentThread().interrupt() to reset it the interruption is lost completely. This means that anything checking Thread.interrupted() or isInterrupted() won’t know that the thread was interrupted which can cause threads to keep running instead of stopping properly during things like job aborts or a graceful shutdown.

I spotted the same issue in doTemplateFileCheck() where both IOException and InterruptedException were caught together in one block again no flag restoration.

So what I did in the fix was split both of those catch blocks: now InterruptedException is handled separately, with the interrupt flag being restored before we render the error while all other exceptions are still handled exactly as before (no change there).

I also added a new test class (EmailExtTemplateActionTest) to make sure the interrupt flag gets correctly restored after calling renderTemplate() with an interrupted thread.

Let me know if you'd like me to tweak anything..

Vinamra-tech:
Great work on catching that, I appreciate it but I do have some concerns about the execution flow. I mean if a thread is interrupted , normally it means the job is being aoborted or rather in worst case system is shutting down. If we go by the pr's proposed changes , we might risk two things. 

1. Doing unnnecessary work on a dying thread.
2. Triggering subsequent ClosedByInterruptExceptions if that rendering process hits any blocking I/O while flag is now set.

I mean , from what I understood from the codebase , shouldn't we just restore the flag and immediately return or fail-fast to respect abort signal.

It would be great if you claified it!!

ArpanC6:
> Great work on catching that, I appreciate it but I do have some concerns about the execution flow. I mean if a thread is interrupted , normally it means the job is being aoborted or rather in worst case system is shutting down. If we go by the pr's proposed changes , we might risk two things.
> 
> 1. Doing unnnecessary work on a dying thread.
> 2. Triggering subsequent ClosedByInterruptExceptions if that rendering process hits any blocking I/O while flag is now set.
> 
> I mean , from what I understood from the codebase , shouldn't we just restore the flag and immediately return or fail-fast to respect abort signal.
> 
> It would be great if you claified it!!

@Vinamra-tech thanks for the detailed feedback these are valid concerns worth discussing
I want to clarify the reasoning behind the current approach. The renderTemplate() method is a @JavaScriptMethod it's called directly from the browser UI when a user wants to preview a template. In this context the "interrupted thread" scenario is different from a background build thread being aborted.

Here's why I chose to render the error instead of returning early:

### 1. User facing method If we return an empty result silently the UI shows nothing and the user has no idea what happened. Rendering the error gives the user visible feedback that something went wrong.


### 2. The interrupt flag IS restored By calling Thread.currentThread().interrupt() before renderError() we correctly restore the flag. Any subsequent blocking I/O inside renderError() would then throw ClosedByInterruptException naturally which is the correct behavior. We're not suppressing the interruption we're just giving the user one last error message before the thread responds to the abort.


### 3. renderError() is lightweight It's just string concatenation no blocking I/O involved. So the risk of ClosedByInterruptException during renderError() itself is essentially zero.
### That said if @slide prefers the fail fast approach with return result I'm happy to update. Just wanted to explain the reasoning behind the current choice.

Vinamra-tech:
Thanks for addressing the query!! @ArpanC6 The context of it being a @JavaScriptMethod is really helpful, and that definitely shifts the perspective.

I'm curious about one edge case regarding the lifecycle here, though. Since this relies on Stapler and Jetty to handle the returned payload, what happens when the framework tries to write that string to the HttpServletResponse output stream?

If we restore the thread's interrupt flag before returning, we're handing an actively interrupted thread back to the servlet container. When Jetty attempts to flush that payload to the network socket, wouldn't hitting that blocking I/O with an active interrupt flag trigger an IOException or ClosedByInterruptException at the container level? If so, the response might be abruptly terminated anyway, meaning the UI still wouldn't get the error message, and we might just end up logging unnecessary framework-level stack traces.

Also, since this is an HTTP worker thread, an interruption usually implies the server is shutting down or the client dropped the connection—in which case the browser might not be waiting for the payload anymore.

I might be overthinking the Jetty lifecycle here, but I'd love to hear your thoughts on whether it's safer to just log a warning and return immediately to avoid container-level exceptions!

Thanks!

ArpanC6:
> Thanks for addressing the query!! @ArpanC6 The context of it being a @JavaScriptMethod is really helpful, and that definitely shifts the perspective.
> 
> I'm curious about one edge case regarding the lifecycle here, though. Since this relies on Stapler and Jetty to handle the returned payload, what happens when the framework tries to write that string to the HttpServletResponse output stream?
> 
> If we restore the thread's interrupt flag before returning, we're handing an actively interrupted thread back to the servlet container. When Jetty attempts to flush that payload to the network socket, wouldn't hitting that blocking I/O with an active interrupt flag trigger an IOException or ClosedByInterruptException at the container level? If so, the response might be abruptly terminated anyway, meaning the UI still wouldn't get the error message, and we might just end up logging unnecessary framework-level stack traces.
> 
> Also, since this is an HTTP worker thread, an interruption usually implies the server is shutting down or the client dropped the connection—in which case the browser might not be waiting for the payload anymore.
> 
> I might be overthinking the Jetty lifecycle here, but I'd love to hear your thoughts on whether it's safer to just log a warning and return immediately to avoid container-level exceptions!
> 
> Thanks!

@Vinamra-tech that's a really sharp observation about the Jetty lifecycle I hadn't considered that layer.
You're right that if the interrupt flag is set when Stapler tries to flush the response the container level I/O could fail regardless meaning the error message might never reach the browser anyway. That does weaken my "user facing feedback" argument.

Given this I think the safer approach is

**} catch (InterruptedException ex) {
    Thread.currentThread().interrupt();
    return result; // restore flag and return immediately
}**

I'll wait for @slide's input before updating happy to make the change if he agrees with this direction.

slide:
Good catch on the interrupt flag — silently swallowing `InterruptedException` without restoring it is a well-known Java anti-pattern and this is worth fixing.

That said, there's a design question from the thread that still needs a decision before this is mergeable:

### Key question: render error or fail-fast?

Since `renderTemplate()` is a `@JavaScriptMethod` running on a Jetty HTTP worker thread, restoring the interrupt flag and then continuing into `renderError()` means we hand an actively interrupted thread back to the servlet container. When Stapler tries to flush the response, that blocking I/O may throw `ClosedByInterruptException` at the container level anyway — meaning the user might not see the error message regardless, and we'd just generate unnecessary framework-level stack traces.

The safer alternative proposed in the thread is:

```java
} catch (InterruptedException ex) {
    Thread.currentThread().interrupt();
    return result; // fail-fast, let the container handle it
}
```

I think the fail-fast approach is probably the best.

### `doTemplateFileCheck()` context

It's also worth confirming whether `doTemplateFileCheck()` is called exclusively from HTTP threads or also from background build threads — the right behaviour may differ depending on the caller, and the fix should reflect that.

### Test coverage for the fail-fast path

If the implementation is updated to return early, the test should also assert that the returned value is an empty/default result rather than an error string, to lock in that behaviour alongside the interrupt flag assertion.

ArpanC6:
@slide Updated the PR with the fail fast approach for renderTemplate() it now restores the interrupt flag and returns immediately instead of calling renderError(). I also updated the test to assert that the returned result is empty/default alongside the interrupt flag check.

Regarding doTemplateFileCheck() from what I see it is called from an HTTP thread (it is a @QueryParameter method handled by Stapler). The same fail fast pattern should apply there as well. Let me know if you would like me to update that too.

Also closed #1571 to avoid overlap.

PR #1554, opened 2026-04-03
[JENKINS-68146] html rename `saveOutput` artifact

### LINKED ISSUE #1378: [JENKINS-68146] html rename saveOutput artifact
<p>When I'm providing saveOutput: true as parameter the output html file is named "Always-10.html" which 10 is the build number.</p>

<p>I'm trying to figure our how to set the output file name to a well known name.</p>

<p> </p>

<p>Consider  I call `emailExt` from `always` block at the end of the pipeline</p>

---
<details><summary><i>Originally reported by <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=barel_elbaz">barel_elbaz</a>, imported from: <a class="original-jira-link" href="https://issues.jenkins.io/browse/JENKINS-68146" target="_blank">html rename saveOutput artifact</a></i></summary>
<i><ul>
<li><b>status</b>: Open
<li><b>priority</b>: Minor
<li><b>votes</b>: 1
<li><b>watchers</b>: 3
</ul></i>

<pre>
<p>When I'm providing saveOutput: true as parameter the output html file is named "Always-10.html" which 10 is the build number.</p>

<p>I'm trying to figure our how to set the output file name to a well known name.</p>

<p> </p>

<p>Consider  I call `emailExt` from `always` block at the end of the pipeline</p></pre>
</details>
</details>

akash-manna-sky:
[JENKINS-68146] html rename `saveOutput` artifact

Fixes #1378 

### Submitter checklist


ArpanC6:
@akash-manna-sky 
This change updates the artifact filename from the dynamic pattern triggername-buildid.[txt|html] to a fixed static name email-ext-message.[txt|html].

For users who rely on the old filename pattern in their pipeline scripts post-build automation or artifact archiving setups (e.g., using archiveArtifacts with a specific glob pattern) this could cause issues.

Has the impact on backward compatibility been taken into account? For example would it be helpful to mention this as a potential breaking change in the PR description or release notes?

akash-manna-sky:
> @akash-manna-sky This change updates the artifact filename from the dynamic pattern triggername-buildid.[txt|html] to a fixed static name email-ext-message.[txt|html].
> 
> For users who rely on the old filename pattern in their pipeline scripts post-build automation or artifact archiving setups (e.g., using archiveArtifacts with a specific glob pattern) this could cause issues.
> 
> Has the impact on backward compatibility been taken into account? For example would it be helpful to mention this as a potential breaking change in the PR description or release notes?

Hi @ArpanC6, thanks for your review note, I really appreciate the effort. But please let the maintainer review the PR first, I would prefer to wait for the maintainer’s review first. Thanks!

slide:
Thanks for linking this to JENKINS-68146 and taking a stab at fixing it. The intent is right, but I have concerns about the approach.

### A static filename isn't the right fix

Replacing the dynamic `triggername-buildid.[txt|html]` pattern with a hardcoded `email-ext-message.[txt|html]` trades one problem for another. Users who want to reliably reference the artifact in downstream scripts now have a predictable name, but they lose any ability to distinguish it from other triggers or customise it to fit their pipeline conventions.

A better approach would be to expose a configurable `saveOutputFileName` parameter (similar to how other options in the plugin work) so users can specify the filename they want. The default could fall back to the current dynamic pattern to preserve backward compatibility. This gives users control without breaking existing setups.

### Static name causes overwrites in multi-trigger builds

If a build fires multiple email triggers, the current dynamic naming produces distinct filenames per trigger. With a fixed static name, later triggers would silently overwrite earlier ones. That's a regression for anyone using multiple triggers per build.

### Backward compatibility

The existing dynamic filename pattern is part of the observable behaviour of the plugin. Any users relying on it in `archiveArtifacts` globs or post-build automation would be broken silently by this change. At minimum this needs to be called out as a breaking change, but ideally the approach above avoids the breakage entirely.

### Testing section is blank

The testing section in the PR description was left as the template placeholder. Please describe how this was tested before merging.

akash-manna-sky:
Still WIP.

akash-manna-sky:
Hi @slide, can you please review this PR? I made all the suggested changes.

PR #1555, opened 2026-04-03
fix: resolve JENKINS-26838 - replace IsChildFileCallable with FilePath.isDescendant()

ArpanC6:
### Summary
This update resolves the long standing TODO JENKINS-26838 in the AbstractEvalContent.java file. It replaces the previous custom implementation of IsChildFileCallable with the newer more efficient FilePath.isDescendant() API, which is now part of Jenkins core.

### Problem
The old implementation of isChildOf() had two main issues:
Symlink Safety: The method used getParent() to traverse directories but it didn't handle symbolic links properly which could lead to incorrect results.
Excessive Boilerplate: The custom logic required a dedicated inner class to do something that could now be done in a single line.

### Solution
By switching to FilePath.isDescendant() the code now benefits from:

Symlink Safety: The new method automatically handles symlinks, ensuring accurate results.
Efficiency: It's a much cleaner solution, reducing unnecessary complexity.
Remote Agent Compatibility: It works seamlessly with remote agents which was a key requirement.

### Changes
The old isChildOf() method was replaced with a call to FilePath.isDescendant().
The IsChildFileCallable inner class was removed as it’s no longer needed.
Unused imports like VirtualChannel and MasterToSlaveFileCallable were also removed.

### Testing
All existing tests ran successfully. The only failures were from the ExtendedEmailPublisherTransientSmtpErrorTest tests which are unrelated to this change and are also failing on the main branch.

### References
JENKINS-26838: https://issues.jenkins.io/browse/JENKINS-26838


ArpanC6:
@slide All CI checks are now passing. Happy to make any changes based on review feedback.

slide:
A test to verify that the change is correct would be good.

ArpanC6:
Thanks for the feedback @slide I have added unit tests for the isChildOf() 
method in AbstractEvalContentTest.java covering the following cases

1) Returns true when the path is a valid child of the parent
2) Returns false when the path is not under the parent directory
3) Returns false when both paths point to the same directory (empty relative path)

These cases also address the coverage warnings that CI reported on the 
partially covered branches.

slide:
The removal of `IsChildFileCallable` is the right direction, and the tests are a good addition. However, looking at the actual implementation, there's a correctness issue that needs to be addressed before this is mergeable.

### The manual `startsWith` check is fragile and introduces a false-positive bug

The new implementation derives a relative path via string prefix matching before calling `isDescendant()`:

```java
if (!childPath.startsWith(parentPath)) {
    return false;
}
String relative = childPath.substring(parentPath.length());
```

String-based prefix matching on paths is not safe. A parent path of `/tmp/parent` would incorrectly match `/tmp/parent-sibling/child` because the string starts with `/tmp/parent`. That path would pass the prefix check, have its "relative" portion computed as `-sibling/child`, and then get passed to `isDescendant()` — likely returning `false` due to the leading `-`, but for the wrong reason and in a fragile way.

There's no test covering this case, which is why it wasn't caught.

### The symlink safety claim isn't fulfilled

The PR description states that switching to `FilePath.isDescendant()` fixes the symlink handling weakness of the old approach. But the manual `startsWith` check runs first and doesn't resolve symlinks, so symlinked paths could still produce incorrect results before `isDescendant()` even gets a chance to evaluate them.

### Suggested fix

The manual prefix check and relative path derivation should be dropped entirely. A simpler and correct approach would be to use `java.nio.file.Path.relativize()` to safely derive the relative path, or to find whether `FilePath` exposes an API that avoids needing the relative path at all. The goal should be to let `isDescendant()` do all the work rather than pre-filtering with string operations.

### Missing test case

Please add a test for the false-positive scenario — e.g. checking `/tmp/parent-sibling/child` against `/tmp/parent` — to ensure this case is caught and returns `false`.

src/main/java/hudson/plugins/emailext/plugins/content/AbstractEvalContent.java
```diff
@@ -155,8 +153,19 @@ protected InputStream getFileInputStream(Run<?, ?> run, FilePath workspace, Stri
     @Restricted(NoExternalUse.class)
     public static boolean isChildOf(final FilePath potentialChild, final FilePath parent)
             throws IOException, InterruptedException {
-        // TODO JENKINS-26838 use API when available in core
-        return parent.act(new IsChildFileCallable(potentialChild));
+        java.nio.file.Path parentNio =
+                java.nio.file.Paths.get(parent.getRemote()).normalize();
```
slide, line 157: Can you import java.nio.file.Paths instead of using the full package?

PR #1557, opened 2026-04-04
Expand tests to verify multipart structure and body content

Faheem12005:
Resolves the TODO in `testShouldSendEmailUsingUtf8ByDefault`

this PR tightens and broadens multipart validation by enforcing expected MIME structure and body-part semantics across relevant tests.
`testShouldSendEmailUsingUtf8ByDefault` previously had a silent fallback for if the message wasn't a `MimeMessage`. I've replaced this with a stricter assertion so potentially broken email implementations don't pass through.
Defined a new helper `assertMultipartSubtype` to verify the multipart subtypes (eg: mixed, alternative) on both the outer message header and the parsed `MimeMultipart` object, and applied it to existing tests to increase multipart coverage.

### Testing done
Modifications to tests only

### Submitter checklist

### DIFF
--- a/src/test/java/hudson/plugins/emailext/ExtendedEmailPublisherTest.java
+++ b/src/test/java/hudson/plugins/emailext/ExtendedEmailPublisherTest.java
@@ -45,8 +45,8 @@ import jakarta.mail.Address;
 import jakarta.mail.BodyPart;
 import jakarta.mail.Message;
 import jakarta.mail.MessagingException;
+import jakarta.mail.Part;
 import jakarta.mail.Session;
-import jakarta.mail.internet.MimeBodyPart;
 import jakarta.mail.internet.MimeMessage;
 import jakarta.mail.internet.MimeMultipart;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


ArpanC6:
@Faheem12005 The findFirstPartByContentType() helper looks good overall but one thing worth considering: the recursive call has no depth limit. For deeply or maliciously nested MimeMultipart structures this could theoretically lead to a stack overflow. Since this is test code it's low risk but adding a simple depth guard (e.g. maxDepth parameter) would make it more robust.

slide:
Good improvement overall — replacing the silent `if (msg instanceof MimeMessage)` fallback with `assertInstanceOf` is exactly the right fix for that TODO, and the `assertMultipartSubtype` helper is well-structured.

A few things worth tidying up:

### `findFirstPartByContentType` is redundant in `testAttachBuildLog`

The helper is called at the end of the test to assert a `text/plain` part exists, but `body.getContentType()` is already asserted to contain `text/plain` just a few lines earlier. The `assertNotNull` on the result of `findFirstPartByContentType` would always pass given the earlier assertions and adds no additional coverage. Either give it a more meaningful use (e.g. a test covering nested multipart structures) or remove the redundant call.

### `findFirstPartByContentType` has no depth limit

As ArpanC6 noted, the recursive call has no guard against deeply nested `MimeMultipart` structures. It's test code so the practical risk is low, but a simple `maxDepth` parameter would be good hygiene. This wasn't addressed after ArpanC6's comment.

### `assertEquals(1, multipart.getCount(), ...)` is a tighter assertion than before

The old code checked `>= 1`, the new code pins it to exactly 1. That's fine if the intent is to lock down the exact structure, but worth confirming this won't become a fragile assertion if the email structure is ever legitimately extended in future (e.g. a footer part).

Faheem12005:
Thanks for the review! Looking into it, I think chaining functions to retrieve nested structures is a better approach than a recursive solution as it makes more sense to follow nesting at a glance, So I've done away with that function. And yes, as for the count assertions I thought pinning the test to the exact expected structure ensures it fails loudly if anything changes rather than silently passing just because an unrelated part happened to satisfy the assertion.

slide (approved):

slide (approved):

PR #1558, opened 2026-04-04
Add Null Checks  and Unit Test for StaticProxyInstanceWhitelist

Vinamra-tech:
Improve error handling and observability in `StaticProxyInstanceWhitelist`

Currently, `getClass().getResource(resource)` returns `null` silently when a 
classpath resource is missing. `StaticWhitelist.from(null)` then throws a 
`NullPointerException` with no indication of which resource was missing or why, 
making it difficult for developers and administrators to diagnose misconfiguration.

This PR adds a defensive null check with a descriptive `IllegalArgumentException` 
naming the missing resource path, and adds `FINE` level logging in `permitsMethod` 
and `permitsFieldGet` to help administrators understand which rules are being 
evaluated — without producing log spam in production (FINE is silent by default 
and can be enabled via Jenkins log configuration when needed).

### Testing done

Added 6 unit tests in `StaticProxyInstanceWhitelistTest` covering:
- Missing single resource throws `IllegalArgumentException` with the resource path in the message
- Missing resource among multiple resources is correctly caught
- `permitsMethod` returns false when no proxy permits the method
- `permitsMethod` returns false when the receiver is not the whitelisted instance
- `permitsFieldGet` returns false when no proxy permits the field
- `permitsFieldSet` always returns false unconditionally

All 6 tests pass locally:
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/groovy/sandbox/StaticProxyInstanceWhitelist.java
+++ b/src/main/java/hudson/plugins/emailext/groovy/sandbox/StaticProxyInstanceWhitelist.java
@@ -5,6 +5,8 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.net.URL;
+import java.util.logging.Logger;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist;
 
@@ -12,6 +14,9 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist;
  * {@link ObjectInstanceWhitelist} backed by a set of {@link StaticWhitelist}s.
  */
 public class StaticProxyInstanceWhitelist extends ObjectInstanceWhitelist<Object> {
+
+    private static final Logger LOGGER = Logger.getLogger(StaticProxyInstanceWhitelist.class.getName());
+
     private Whitelist[] proxies;
 
     public StaticProxyInstanceWhitelist(Object instance, @NonNull String... resources) throws IOException {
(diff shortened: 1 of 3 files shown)


ArpanC6:
@Vinamra-tech  Good catch on the null check. One thing I noticed the original line proxies[i] = StaticWhitelist.from(getClass().getResource(resource)); is still present before the null check. This means getResource() is called twice and proxies[i] gets assigned once without the null guard before being overwritten by the second assignment. The first line should be removed so only the null checked assignment remains.

Vinamra-tech:
> @Vinamra-tech Good catch on the null check. One thing I noticed the original line proxies[i] = StaticWhitelist.from(getClass().getResource(resource)); is still present before the null check. This means getResource() is called twice and proxies[i] gets assigned once without the null guard before being overwritten by the second assignment. The first line should be removed so only the null checked assignment remains.

Thanks for taking the time to review! I think there may be some confusion — the diff actually shows the original proxies[i] = StaticWhitelist.from(getClass().getResource(resource)) line was replaced (shown as a removal in the diff), not duplicated. getResource() is now called exactly once, stored in url, null-checked, and then passed to StaticWhitelist.from(url).

slide:
Clean fix — failing fast with a descriptive `IllegalArgumentException` is much better than a cryptic NPE from `StaticWhitelist.from(null)`, and the lambda-based `LOGGER.fine()` calls are the right approach to avoid unnecessary string concatenation when logging is disabled.

A couple of minor things worth addressing:

### ArpanC6's concern is already resolved

Just to close the loop: the diff clearly shows the original `proxies[i] = StaticWhitelist.from(getClass().getResource(resource))` line was removed and replaced. `getResource()` is called exactly once, stored in `url`, null-checked, then passed to `StaticWhitelist.from(url)`. The author's response is correct — no change needed here.

### `constructorThrowsOnOneOfMultipleMissingResources` doesn't test the intended scenario

The test passes two missing resources, so the loop fails on the first one immediately. It doesn't actually verify the behaviour when the first resource is valid and a subsequent one is missing — which is the more interesting case. Consider passing a real classpath resource first followed by a missing one, to confirm the loop correctly catches failures beyond the first iteration.

### `String.class.getDeclaredField("value")` is fragile

The field tests reference an internal JDK field that may not exist or be accessible across all JVM implementations or future Java versions. Using a field from a test-owned class or a stable public field from the standard library would be more robust.

Vinamra-tech:
@slide Sure thing
1. The test now passes a real classpath resource (test-whitelist.txt, placed at src/test/resources/hudson/plugins/emailext/groovy/sandbox/test-whitelist.txt so getClass().getResource() resolves it correctly relative to the package) as the first argument, followed by the missing one. This confirms the loop iterates past a valid resource and correctly catches the failure on the second iteration. All 6 tests pass locally.

2.On String.class.getDeclaredField("value") — agreed, that was fragile. Replaced with a private static inner class FieldHolder with a public String name field, owned entirely by the test. No JDK internals involved, no accessibility workarounds needed.

PR #1559, opened 2026-04-05
test: add unit tests for EmailThrottler

ArpanC6:
### Summary
`EmailThrottler` is a singleton class responsible for rate limiting 
email sending in the plugin. Despite being a core utility class it 
had no test coverage. This PR adds unit tests to verify its behavior.

### Problem
Without tests, there is no automated verification that:
- The throttling limit works correctly
- The reset mechanism functions as expected
- The singleton pattern is properly enforced

Any future change to `EmailThrottler` could silently break these 
behaviors without being caught by CI.

### Changes
Added `EmailThrottlerTest.java` covering the following cases:
- Throttling limit is not exceeded initially (fresh instance)
- Throttling limit is exceeded after incrementing to the limit
- Reset correctly clears the email count
- Singleton pattern returns the same instance across calls

### Testing
All 4 tests passing locally. No functional changes test only.

### DIFF
--- a/src/test/java/hudson/plugins/emailext/EmailThrottlerTest.java
+++ b/src/test/java/hudson/plugins/emailext/EmailThrottlerTest.java
@@ -1,0 +1,50 @@
+package hudson.plugins.emailext;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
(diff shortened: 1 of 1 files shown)


ArpanC6:
@slide All CI checks are now passing. Happy to make any changes based on 
review feedback.

While working on this I had a couple of questions

1) Would it be cleaner to express THROTTLING_PERIOD using 
TimeUnit.HOURS.toNanos(1) instead of 60 * 60 * 1000 * 1000000L 
for better readability?

2) Since EmailThrottler uses System.nanoTime() for period tracking 
would the throttle period reset unexpectedly if Jenkins restarts 
mid period? Was this a conscious design decision?

PR #1560, opened 2026-04-05
Fix type inference issue in attachBuildLog

Eng-YasminKotb:
### Description of changes

This PR fixes a type inference issue in `AttachmentUtils.attachBuildLog`.

Using `var` for the `all` variable caused incompatibility with the more specific generic type returned by `MatrixAssist#getExactRuns`.  
Replacing it with an explicit type (`List<? extends Run<?, ?>>`) resolves the issue and improves type safety.

### Testing done

Added a simple test to verify that build logs are correctly attached when `attachBuildLog` is enabled.

### Submitter checklist
- [x] Make sure you are opening from a topic/feature/bugfix branch
- [ ] Link to relevant pull requests
- [x] Ensure you have provided tests that demonstrate the fix works

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -63,7 +63,7 @@ In the *Post-build Actions* section, click on *Add post-build action* and then s
 
 There are three main fields that you can edit when this plugin is enabled:
 
-Project Recipient List:: This is a comma (or whitespace) separated list of email recipients. Allows you to specify a single recipient list for each email that is sent.
+Project Recipient List:: This is a comma (or whitespace) separated list of email recipients. It allows you to specify a single recipient list for each email that is sent.
 Default Subject:: This allows you to configure a token (more about tokens later) that can be used to easily configure all email subjects for the project.
 Default Content:: Same as *Default Subject*, but for the email body instead of the subject.
 
@@ -124,12 +124,12 @@ To see what conditions must be met for this plugin to send an email, click on th
 
 ==== Extended Email Publisher Triggers
 
-The following triggers are available as part of the Extended Email Publisher plugin, other plugins can provide triggers as well through the extension point defined in the Extended Email Publisher:
+The following triggers are available as part of the Extended Email Publisher plugin, other plugins can provide triggers as well, through the extension point defined in the Extended Email Publisher:
 
 Aborted:: An email will be sent if the build status is "Aborted". A build is aborted via the UI or API, but normally requires some sort of user intervention to occur. An aborted build is stopped during its execution.
 Always:: Always triggers an email after the build, regardless of the status of the build.
 Before Build:: An email will be sent when the build begins, but after SCM polling has completed.
-Failure -> Unstable (Test Failures):: An email will be sent any time the build goes from failing (compilation or build step failures), to unstable (unit test failures). This basically means that all the builds steps were successful, but there are still tests failing.
+Failure -> Unstable (Test Failures):: An email will be sent any time the build goes from failing (compilation or build step failures) to unstable (unit test failures). This basically means that all the builds steps were successful, but there are still tests failing.
 Failure - Any:: An email will be sent any time the build fails.  If the "Failure - Still" trigger is configured, and the previous build status was "Failure", then the "Failure - Still" trigger will send an email instead.
 Failure - 1st:: An email will be sent when the build status changes from "Success" to "Failure".
 Failure - 2nd:: An email will be sent when the build fails twice in a row after a successful build.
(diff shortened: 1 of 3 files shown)


Eng-YasminKotb:
It looks like the CI failure is related to infrastructure (pr-head not building and Windows bat step), as tests are passing. Retrying the build.

Eng-YasminKotb:
Hi, @slide!

I've fixed the issues and updated the PR.
Everything should be good now.

Would appreciate your review when you have time. Thanks!

ArpanC6:
@Eng-YasminKotb The production code fix looks good replacing var with the explicit List<? extends Run<?, ?>> solves the type inference issue as expected. Just a couple of small style suggestions: I noticed that the test file uses wildcard imports (import static org.junit.jupiter.api.Assertions.* and import hudson.model.*). These are generally avoided in our codebase and we prefer using explicit imports instead. Also I see there are several CI retry commits in the history. It would be cleaner to squash those commits before merging.

src/test/java/hudson/plugins/emailext/AttachmentUtilsTest.java
```diff
@@ -12,17 +12,12 @@
 import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertInstanceOf;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
slide, line 15: * imports will not be accepted

src/test/java/hudson/plugins/emailext/AttachmentUtilsTest.java
```diff
@@ -12,17 +12,12 @@
...
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.AbstractBuild;
-import hudson.model.BuildListener;
-import hudson.model.FreeStyleBuild;
-import hudson.model.FreeStyleProject;
+import hudson.model.*;
```
slide, line 20: No * imports please

README.adoc
slide, line 66: Looks like you have multiple PR's worth of changes in this PR. If you want to have separate PR's you need to use completely separate branches that you are pushing to.

slide:
What does this actually fix? Is it an actual runtime issue? What cases will this actually fail in an actual Jenkins instance?

Eng-YasminKotb:
Thanks, @slide for the feedback!

this change is more of a type-safety improvement rather than a runtime bug. The code was working, but using `var` here led to less clear type inference, so I replaced it with an explicit type.

I also see your point about having multiple changes in one PR. My main intention was the change in AttachmentUtils, and I added the test just to verify the behavior. I can split them into separate PRs if that would be preferred.

I'll also clean up the code and address the comments.

README.adoc
Eng-YasminKotb: I realized that my branch was not up to date with main, which caused previously merged changes (like the README update) to appear in this PR.

I'll rebase and clean up the branch to keep the PR focused only on the intended change.

Eng-YasminKotb:
Thanks ,@ArpanC6!

You're right about the wildcard imports, I’ll replace them with explicit imports to follow the project conventions.

I’ll also clean up the commit history and squash the CI retry commits.

Eng-YasminKotb:
Hi, @slide 
I’ve updated the changes and addressed the comments.
Do you prefer I open a new PR with a clean branch(I have already created it), or should I keep this one and proceed with the current changes?

slide:
I would like a clean PR that addresses only the items in the title of the PR. It looks like there are still star imports, there are still unrelated changes in the README

The test uses a hand-rolled loop with a boolean flag to check for an attachment named `log`, then calls `assertTrue(hasLog)` with no message. This could be simplified and made more readable, and the assertion should have a descriptive failure message. It's also worth checking whether `testAttachBuildLogCompress` already covers similar ground before adding a new test.

src/main/java/hudson/plugins/emailext/AttachmentUtils.java
```diff
@@ -274,11 +254,12 @@ private static void attachSingleLog(
 
     public static void attachBuildLog(ExtendedEmailPublisherContext context, Multipart multipart, boolean compress) {
         var main = context.getRun();
```
slide, line 256: Wouldn't we want to replace `var` on main as well?

Eng-YasminKotb: Since my last update, I had already prepared a separate branch for a cleaner version of the change.

I'll close this PR and open a new one focused solely on the type inference fix in attachBuildLog. After reviewing the existing tests, I agree that the new test does not add meaningful coverage here.

Regarding your comment about main, my understanding is that the type incompatibility comes from `var` all, since it is reassigned with the value returned by getExactRuns, while main is already inferred as `Run<?, ?>` from context.getRun().
For the new PR, would you prefer making `main` explicit as well for consistency/readability, or keeping the change limited to the variables involved in the type inference issue, @slide?

slide: I would like to change `main` to an explicit type for consistency. Thanks!

PR #1561, opened 2026-04-06
fix: restore interrupt flag in ContentBuilder.transformText() and add test coverage

ArpanC6:
What was wrong?

While going through the codebase I noticed an issue in the ContentBuilder.transformText() method the key method used to expand all email tokens during Jenkins build notifications. Specifically there was a broad catch (Exception e) block that silently swallowed InterruptedException without properly restoring the thread's interrupt flag.

In Java when an InterruptedException is caught the JVM automatically clears the interrupt flag. If the code doesn't explicitly call Thread.currentThread().interrupt() to restore that flag the interruption is effectively lost. This means any thread that checks Thread.interrupted() or isInterrupted() won’t know that an interruption happened.

This is a big deal in this case because transformText() is called every time an email is sent. So if a build thread gets interrupted during the token expansion (for example, during a job abort or a graceful Jenkins shutdown) it silently loses its interrupt status and keeps running instead of stopping cleanly.

The root of the issue lies in TokenMacro.expandAll() which declares throws InterruptedException. However the existing catch (Exception e) block was treating it the same as any other exception logging it as SEVERE and continuing without restoring the interrupt flag.

What I changed

To fix this I made some adjustments in ContentBuilder.java. I split the broad catch (Exception e) block into two separate blocks:

catch (InterruptedException e): This specifically handles interruptions. It now restores the interrupt flag by calling Thread.currentThread().interrupt() and logs it at a WARNING level since interruptions are expected signals not unexpected errors.
catch (Exception e): This block still handles all other exceptions as before without any change in behavior for non interrupt cases.
Test added

I also added a test to make sure the interrupt flag gets properly restored. The test testTransformText_restoresInterruptFlagWhenInterrupted() is now part of the ContentBuilderTest class. It does the following:

Sets the thread's interrupt flag before calling transformText().
Calls transformText() with a valid token.
Verifies that the interrupt flag is correctly restored once the method returns.

This ensures that the fix works as expected and that future changes to transformText() won’t accidentally reintroduce this problem.

Result:
Tests run: 11
Failures: 0
Errors: 0

The fix is in place and everything’s working smoothly.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
@@ -92,6 +92,9 @@ public final class ContentBuilder {
             }
         } catch (MacroEvaluationException e) {
             context.getListener().getLogger().println("Error evaluating token: " + e.getMessage());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            Logger.getLogger(ContentBuilder.class.getName()).log(Level.WARNING, "Token expansion interrupted", e);
         } catch (Exception e) {
             Logger.getLogger(ContentBuilder.class.getName()).log(Level.SEVERE, null, e);
         }
--- a/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
@@ -94,7 +94,7 @@ public final class ContentBuilder {
             context.getListener().getLogger().println("Error evaluating token: " + e.getMessage());
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
-            Logger.getLogger(ContentBuilder.class.getName()).log(Level.WARNING, "Token expansion interrupted", e);
+            Logger.getLogger(ContentBuilder.class.getName()).log(Level.INFO, "Token expansion interrupted", e);
         } catch (Exception e) {
             Logger.getLogger(ContentBuilder.class.getName()).log(Level.SEVERE, null, e);
         }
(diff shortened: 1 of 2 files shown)


ArpanC6:
@slide just wanted to quickly flag this. While working on fixing the interrupt flag in EmailExtTemplateAction I came across the same issue in ContentBuilder.transformText(). This one’s a bit more critical since transformText() is the core method that's called every time an email is sent for build notifications. If a build thread gets interrupted here during token expansion the interrupt flag is lost silently and the thread just keeps going instead of stopping properly.

I’ve split the catch block restored the interrupt flag and added a test to make sure it behaves as expected. Let me know if you’d like any tweaks or further changes.

ArpanC6:
@slide 
Note: The code coverage drop shown here reflects the existing low coverage on ContentBuilder this PR adds one new test (testTransformText_restoresInterruptFlagWhenInterrupted) specifically to verify the interrupt flag fix. Happy to add more tests for ContentBuilder in a follow up if that would be helpful.

slide:
Unlike #1553, this one doesn't have the Jetty/Stapler concern — `transformText()` runs on a build thread, so restoring the interrupt flag and letting the caller handle it is the correct approach here. The fix is minimal and well-placed.

Two things worth tidying up:

### The test consumes the interrupt flag it's trying to verify

`Thread.interrupted()` is a clearing call — it returns `true` and then clears the flag. This means if anything runs after this test on the same thread, it won't see a stale interrupt flag, but it also means the assertion is consuming the very state it's checking. Prefer `Thread.currentThread().isInterrupted()` for the assertion, and then explicitly clear the flag in an `@AfterEach` or `finally` block to ensure test isolation:

```java
assertTrue(Thread.currentThread().isInterrupted(), "Interrupt flag should be restored after InterruptedException");
// clear the flag so it doesn't bleed into subsequent tests
Thread.interrupted();
```

### `WARNING` is too loud for an expected signal

Thread interruption during a build is a normal control flow event — it happens during job aborts and graceful shutdowns. Logging it at `WARNING` could alarm administrators who see it in their logs without context. `INFO` or `FINE` would be more appropriate and consistent with how Jenkins itself typically handles interruptions.

PR #1562, opened 2026-04-07
fix: restore interrupt flag in addContent() save email output path

ArpanC6:
### What was wrong
In ExtendedEmailPublisher.addContent() there was an issue with how the method handles exceptions. The block that saves the email output catches both IOException and InterruptedException together. However when an InterruptedException occurs the thread's interrupt flag wasn’t being restored. This is a problem because if the interrupt flag is lost it can affect the job’s behavior during aborts or graceful shutdowns in Jenkins.

The way it was written the interrupt flag would silently be cleared which is not ideal.

### What I changed
I split the combined catch (IOException | InterruptedException e) block into two separate blocks

1) catch (InterruptedException e) Here I restored the interrupt flag by calling Thread.currentThread().interrupt() before logging the error.
2) catch (IOException e) This part remains unchanged still handling I/O errors as it did before.

### Test
All existing 66 tests in ExtendedEmailPublisherTest ran successfully without any failures.

Tests run - 66
Failures - 0
Errors - 0


ArpanC6:
@slide 
Fix for InterruptedException handling
Split the combined catch block into separate IOException and InterruptedException blocks. The InterruptedException block now restores the interrupt flag using Thread.currentThread().interrupt() before logging. No change in behavior for IOException.

All existing tests pass successfully (66 tests).

slide:
Correct fix — `addContent()` runs on a build thread so restoring the interrupt flag and continuing is the right approach, same as #1561.

A couple of things worth addressing:

### No test added

The sibling PR #1561 included a test verifying interrupt flag restoration. It would be consistent to add one here too — set the interrupt flag before calling `addContent()`, verify it's restored afterward, and clear it in a `finally` block or `@AfterEach` to avoid bleeding into subsequent tests.

### The `InterruptedException` log message will print `null`

Both catch blocks use the same message:
```java
"Error trying to save email output to file. " + e.getMessage()
```

For `InterruptedException`, `e.getMessage()` typically returns `null`, so operators will see `"Error trying to save email output to file. null"` in the build console — which is confusing. The interrupt case should have its own message that gives meaningful context, for example:

```java
"Email output save interrupted; build may be aborting."
```

PR #1565, opened 2026-04-09
Title idea: Exception handling improvement in RecipientProviderUtilities

ArpanC6:
### What was wrong

In the method getByLegacyUserCause() reflection was used to access the authenticationName field. Two things could go wrong here 

1) getDeclaredField() might throw a NoSuchFieldException if the field doesn't exist.

2) authenticationName.get() might throw an IllegalAccessException if access is denied.

Both were caught in a broad catch (Exception e) block making it unclear from the logs which of the two failures actually happened.

In getByUserIdCause() a broad catch (Exception e) was hiding a specific IllegalStateException that came from run.getCause() when the TransientActionFactory Cache wasn't initialized correctly.

In RequesterRecipientProvider.addRecipients() SpotBugs detected a potential null dereference. The method loops through builds using a while loop. If cur becomes null the loop breaks but then it was still passing null to addUserTriggeringTheBuild() which could have caused issues.

### What I changed

1) In getByLegacyUserCause()
I split the broad catch block into specific catches for NoSuchFieldException , IllegalAccessException and IllegalStateException. Each one now has its own warning message so the logs are much clearer about what actually failed.

2) In getByUserIdCause()
I replaced the broad catch with a more specific IllegalStateException catch so now it properly logs that error and doesn't swallow other important exceptions.

2) In RequesterRecipientProvider.addRecipients()
I added a null check before calling addUserTriggeringTheBuild() which prevents the null dereference issue. This should stop potential NPEs (Null Pointer Exceptions).

### Why this is better

1) The logs now clearly show exactly what went wrong so debugging is easier and more transparent.

2) The potential null dereference bug is fixed meaning no more NPEs when cur is null.

3) Instead of catching every possible exception we're now handling specific exceptions making the code more precise and the error handling clearer.

### Tests

Tests run - 1
Failures - 0
Errors - 0

All CI checks passed successfully.


ArpanC6:
@slide
All CI checks are green now

This PR addresses three issues in the recipients package

1) In RecipientProviderUtilities - I replaced the broad exception handling with specific catches. This means that the logs will now clearly show exactly what failed during reflection or when the cache was initializing. It makes troubleshooting way easier.

2) In RequesterRecipientProvider - I fixed a null dereference that SpotBugs had flagged. The issue was that cur was being passed as null to addUserTriggeringTheBuild() after the loop broke early. Now we make sure cur isn’t null before calling that method.

I’m happy to make any changes if needed. Let me know what you think.

One thing I wanted to ask - Since getByLegacyUserCause() is marked as @Deprecated do you think it’s still worth keeping the specific exception handling in there? Or is there a plan to remove this method entirely in a future release?

slide:
The null guard in `RequesterRecipientProvider` is a genuine bug fix and the specific exception catches are a clear improvement over the broad `catch (Exception e)` blocks. A few things worth considering:

### `getByUserIdCause` may now let unexpected exceptions propagate

The original broad catch was a safety net — anything unexpected inside the try block was caught and logged. The new `catch (IllegalStateException e)` only handles that one case. If anything else throws unexpectedly, it will now propagate up to the caller unhandled. Since this runs during recipient resolution for email sending, an uncaught exception could break the entire notification. Consider keeping a fallback `catch (Exception e)` for truly unexpected cases, or confirm that the try block can only ever throw `IllegalStateException`.

### Log level for `NoSuchFieldException` may be too loud

`getByLegacyUserCause` is `@Deprecated` and uses reflection to access a field that may not exist in all Jenkins versions — a `NoSuchFieldException` here is an expected miss, not an error condition. Logging it at `WARNING` could produce noise in every Jenkins instance that hits this path. `FINE` or `INFO` would be more appropriate for that specific case.

### No new tests for the null guard fix

The PR reports 1 test passing but doesn't appear to add a new one specifically for the `cur == null` path. Given this is a real bug fix rather than just a logging improvement, a test covering the case where the upstream chain resolves to `null` before `addUserTriggeringTheBuild` is called would be valuable — similar to the pattern in #1539's `testBrokenUpstreamProjectLinkag`.

PR #1568, opened 2026-04-12
Cache EnvVars in ThreadLocal during sandboxed script/template execution

Vinamra-tech:
Caches `EnvVars` in a `ThreadLocal` for the duration of a single sandboxed 
script/template execution to avoid redundant `build.getEnvironment(listener)` 
calls on every variable lookup.

Previously, `getEnvironment()` was called on every variable reference within 
a sandboxed Groovy email template — once in `permitsMethod()` and again in 
`methodMissing()`. This is an expensive I/O and CPU operation (merging node 
vars, job properties, parameters, etc.). In large templates with many 
variables, this results in significant redundant processing.

**Changes:**
- `EmailExtScriptTokenMacroWhitelist` — added `static final ThreadLocal<EnvVars>`, 
  `beginExecution()`, `endExecution()`, and cache-first lookup in `permitsMethod()`
- `EmailExtScript` — cache-first lookup in `methodMissing()` with fallback
- `ScriptContent` — `beginExecution/endExecution` wrapped in `try-finally` 
  around both sandbox execution points (`renderTemplate` and `executeScript`)
- `EmailExtScriptTokenMacroWhiteListTest` — new test class (JUnit 5)

The cache lives purely in Java (not in the Groovy `Binding`), so scripts 
cannot touch or mutate it. Cleanup is guaranteed via `finally`, preventing 
any cross-execution leakage or memory retention.

Closes #1548
Follows up on #1542 (closed)

### Testing done

Added `EmailExtScriptTokenMacroWhiteListTest`  covering:
- `testGetEnvironmentCalledOnlyOnce` — verifies that 
  `getEnvironment()` is called exactly once across multiple cache lookups
- `testCacheIsClearedAfterEndExecution` — verifies `ENV_CACHE.get()` 
  returns `null` after `endExecution()` is called
- `testFallbackWhenCacheNotInitialized` — verifies graceful behavior 
  when `beginExecution()` was never called
- `testProfilingGain` (`@Tag("performance")`) — benchmarks baseline vs 
  cached execution over 100 macro evaluations with 5 JVM warm-up runs

Profiling results (100 macro evaluations, 5ms simulated I/O per 
`getEnvironment()` call, 5 JVM warm-up runs discarded):

| Scenario            | Time   |
|---------------------|--------|
| Baseline (no cache) | 520 ms |
| ThreadLocal cache   | 5 ms   |
| Reduction           | ~99%   |

All 4 tests pass (`BUILD SUCCESS`).

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,6 +260,7 @@
             <mail.pop3.class>org.jvnet.mock_javamail.MockStore</mail.pop3.class>
             <mail.imap.class>org.jvnet.mock_javamail.MockStore</mail.imap.class>
           </systemPropertyVariables>
+          <excludedGroups>performance</excludedGroups>
         </configuration>
       </plugin>
     </plugins>
--- a/src/main/java/hudson/plugins/emailext/groovy/sandbox/EmailExtScriptTokenMacroWhitelist.java
+++ b/src/main/java/hudson/plugins/emailext/groovy/sandbox/EmailExtScriptTokenMacroWhitelist.java
@@ -43,25 +43,52 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.AbstractWhitelist
 import org.jenkinsci.plugins.tokenmacro.TokenMacro;
 
 /**
- * {@link Whitelist} for the {@link org.jenkinsci.plugins.tokenmacro.TokenMacro} expansion in {@link hudson.plugins.emailext.plugins.content.EmailExtScript}.
+ * {@link Whitelist} for the {@link org.jenkinsci.plugins.tokenmacro.TokenMacro} expansion
+ * in {@link hudson.plugins.emailext.plugins.content.EmailExtScript}.
  */
 public class EmailExtScriptTokenMacroWhitelist extends AbstractWhitelist {
+
+    private static final Logger LOGGER = Logger.getLogger(EmailExtScriptTokenMacroWhitelist.class.getName());
+
+    /**
+     * Caches the EnvVars for the duration of a single script execution.
+     * Populated via beginExecution() and cleared via endExecution() in a
+     * try-finally block in ScriptContent, ensuring no memory leak and no
+     * script-side mutation (lives purely in Java, not in the Groovy Binding).
+     */
+    public static final ThreadLocal<EnvVars> ENV_CACHE = new ThreadLocal<>();
 
     private final List<TokenMacro> macros;
 
     public EmailExtScriptTokenMacroWhitelist() {
         List<TokenMacro> list = new ArrayList<>();
-
         list.addAll(TokenMacro.all());
         list.addAll(ContentBuilder.getPrivateMacros());
-
         this.macros = Collections.unmodifiableList(list);
+    }
+
+    /**
+     * Must be called once before the sandboxed script runs.
+     * Eagerly resolves and caches the EnvVars for this execution.
+     */
+    public static void beginExecution(Run<?, ?> build, TaskListener listener) {
+        try {
+            ENV_CACHE.set(build.getEnvironment(listener));
+        } catch (IOException | InterruptedException e) {
+            LOGGER.log(Level.WARNING, e, () -> "Failed to pre-load environment for " + build.getExternalizableId());
+        }
+    }
+
+    /**
+     * Must be called in a finally block after the sandboxed script finishes.
+     * Clears the cache to prevent any cross-execution leakage.
+     */
+    public static void endExecution() {
+        ENV_CACHE.remove();
     }
 
     @Override
     public boolean permitsMethod(@NonNull Method method, @NonNull Object receiver, @NonNull Object[] args) {
-        // method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object (SimpleTemplateScript2
-        // BUILD_ID)
         if (method.getDeclaringClass() == GroovyObject.class
                 && receiver instanceof EmailExtScript script
                 && "invokeMethod".equals(method.getName())
@@ -73,19 +100,24 @@ public class EmailExtScriptTokenMacroWhitelist extends AbstractWhitelist {
                     return true;
                 }
             }
-            // Else check environment
+
+            // Check environment — use cached EnvVars if available
+            EnvVars vars = ENV_CACHE.get();
+            if (vars != null) {
+                return vars.containsKey(name);
+            }
+
+            // Fallback: no cache available, resolve directly
             Run<?, ?> build = (Run<?, ?>) script.getBinding().getVariable("build");
             TaskListener listener = (TaskListener) script.getBinding().getVariable("listener");
             try {
-                EnvVars vars = build.getEnvironment(listener);
-                return vars.containsKey(name);
+                return build.getEnvironment(listener).containsKey(name);
             } catch (IOException | InterruptedException e) {
-                Logger.getLogger(getClass().getName())
-                        .log(
-                                Level.WARNING,
-                                e,
-                                () -> "Failed to expand environment when evaluating " + name + " on "
-                                        + build.getExternalizableId());
+                LOGGER.log(
+                        Level.WARNING,
+                        e,
+                        () -> "Failed to expand environment when evaluating " + name + " on "
+                                + build.getExternalizableId());
             }
         }
         return false;
--- a/src/main/java/hudson/plugins/emailext/groovy/sandbox/EmailExtScriptTokenMacroWhitelist.java
+++ b/src/main/java/hudson/plugins/emailext/groovy/sandbox/EmailExtScriptTokenMacroWhitelist.java
@@ -56,7 +56,7 @@ public class EmailExtScriptTokenMacroWhitelist extends AbstractWhitelist {
      * try-finally block in ScriptContent, ensuring no memory leak and no
      * script-side mutation (lives purely in Java, not in the Groovy Binding).
      */
-    public static final ThreadLocal<EnvVars> ENV_CACHE = new ThreadLocal<>();
+    private static final ThreadLocal<EnvVars> ENV_CACHE = new ThreadLocal<>();
 
     private final List<TokenMacro> macros;
 
@@ -65,6 +65,10 @@ public class EmailExtScriptTokenMacroWhitelist extends AbstractWhitelist {
         list.addAll(TokenMacro.all());
         list.addAll(ContentBuilder.getPrivateMacros());
         this.macros = Collections.unmodifiableList(list);
+    }
+
+    public static EnvVars getCachedEnvVars() {
+        return ENV_CACHE.get();
     }
 
     /**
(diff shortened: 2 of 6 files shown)


ArpanC6:
I really like this optimization. The way you use ThreadLocal caching is really clean. The try finally guarantee in ScriptContent is great because it makes sure there is no cross execution leakage.

I have a thoughts on this

1. The ENV_CACHE is public now. I think this is a problem because scripts run in a sandbox. If we expose the ThreadLocal directly it feels like a risk. Would it be better to make it private and only expose beginExecution and endExecution? This way other code cannot. Read the cache by accident when it is not supposed to.

2. The testProfilingGain function does not actually check if anything is true. It just prints the results. If the cache stops working for some reason this test will still pass. I think it would be better if we added a threshold assertion. For example we could say that cachedMs has to be less than baselineMs divided by 10. This way the test will actually make sure the cache is working.

3. There is one path in the beginExecution function that is not covered by tests. This is the path where an IOException or InterruptedException happens. The CI coverage report also says that lines 77-78 are not covered. If we add a test that simulates getEnvironment throwing an exception we can cover this path.

4. This is a thing but other test classes, in this code use the @DisplayName annotation to make them easier to read. It might be an idea to add this here too just to be consistent.

Overall I think the main idea of this optimization is really good and the performance numbers are very impressive.

Vinamra-tech:
> I really like this optimization. The way you use ThreadLocal caching is really clean. The try finally guarantee in ScriptContent is great because it makes sure there is no cross execution leakage.
> 
> I have a thoughts on this
> 
> 1. The ENV_CACHE is public now. I think this is a problem because scripts run in a sandbox. If we expose the ThreadLocal directly it feels like a risk. Would it be better to make it private and only expose beginExecution and endExecution? This way other code cannot. Read the cache by accident when it is not supposed to.
> 2. The testProfilingGain function does not actually check if anything is true. It just prints the results. If the cache stops working for some reason this test will still pass. I think it would be better if we added a threshold assertion. For example we could say that cachedMs has to be less than baselineMs divided by 10. This way the test will actually make sure the cache is working.
> 3. There is one path in the beginExecution function that is not covered by tests. This is the path where an IOException or InterruptedException happens. The CI coverage report also says that lines 77-78 are not covered. If we add a test that simulates getEnvironment throwing an exception we can cover this path.
> 4. This is a thing but other test classes, in this code use the @DisplayName annotation to make them easier to read. It might be an idea to add this here too just to be consistent.
> 
> Overall I think the main idea of this optimization is really good and the performance numbers are very impressive.

Hey @ArpanC6, thanks for the review! Addressing each point:

**1. ENV_CACHE visibility — Valid, fixed.**
Agreed completely. ENV_CACHE is now private with a package-private 
getCachedEnvVars() accessor. No external code can reach the ThreadLocal 
directly anymore.

**2. Threshold assertion in testProfilingGain — Respectfully disagree.**
The test is intentionally tagged @Tag("performance") and kept 
assertion-free. Adding a hard threshold like cachedMs < baselineMs / 10 
would make it flaky on shared CI nodes where GC pauses, CPU contention, 
and JVM warm-up variance can easily skew timing results. The purpose of 
this test is local observability during profiling runs, not a correctness 
gate — that role is already covered by testGetEnvironmentCalledOnlyOnce 
which verifies via Mockito that getEnvironment() is called exactly once. 
If the cache stops working, that test fails. The profiling test doesn't 
need to duplicate that responsibility.

**3. Exception path coverage — Valid, fixed.**
Added testBeginExecutionHandlesIOException and 
testBeginExecutionHandlesInterruptedException. Both verify that the cache 
remains null when getEnvironment() throws, confirming the fallback in 
permitsMethod() will kick in gracefully. All 6 tests pass.

**4. @DisplayName — Can't find precedent in the codebase.**
I went through the existing test classes and none of them use 
@DisplayName. Adding it here would introduce inconsistency rather than 
improve it. If there's a specific test class you're referencing that uses 
it, happy to follow that pattern — but I'd rather stay consistent with 
what's already there.

ArpanC6:
> > I really like this optimization. The way you use ThreadLocal caching is really clean. The try finally guarantee in ScriptContent is great because it makes sure there is no cross execution leakage.
> > I have a thoughts on this
> > 
> > 1. The ENV_CACHE is public now. I think this is a problem because scripts run in a sandbox. If we expose the ThreadLocal directly it feels like a risk. Would it be better to make it private and only expose beginExecution and endExecution? This way other code cannot. Read the cache by accident when it is not supposed to.
> > 2. The testProfilingGain function does not actually check if anything is true. It just prints the results. If the cache stops working for some reason this test will still pass. I think it would be better if we added a threshold assertion. For example we could say that cachedMs has to be less than baselineMs divided by 10. This way the test will actually make sure the cache is working.
> > 3. There is one path in the beginExecution function that is not covered by tests. This is the path where an IOException or InterruptedException happens. The CI coverage report also says that lines 77-78 are not covered. If we add a test that simulates getEnvironment throwing an exception we can cover this path.
> > 4. This is a thing but other test classes, in this code use the @DisplayName annotation to make them easier to read. It might be an idea to add this here too just to be consistent.
> > 
> > Overall I think the main idea of this optimization is really good and the performance numbers are very impressive.
> 
> Hey @ArpanC6, thanks for the review! Addressing each point:
> 
> **1. ENV_CACHE visibility — Valid, fixed.** Agreed completely. ENV_CACHE is now private with a package-private getCachedEnvVars() accessor. No external code can reach the ThreadLocal directly anymore.
> 
> **2. Threshold assertion in testProfilingGain — Respectfully disagree.** The test is intentionally tagged @tag("performance") and kept assertion-free. Adding a hard threshold like cachedMs < baselineMs / 10 would make it flaky on shared CI nodes where GC pauses, CPU contention, and JVM warm-up variance can easily skew timing results. The purpose of this test is local observability during profiling runs, not a correctness gate — that role is already covered by testGetEnvironmentCalledOnlyOnce which verifies via Mockito that getEnvironment() is called exactly once. If the cache stops working, that test fails. The profiling test doesn't need to duplicate that responsibility.
> 
> **3. Exception path coverage — Valid, fixed.** Added testBeginExecutionHandlesIOException and testBeginExecutionHandlesInterruptedException. Both verify that the cache remains null when getEnvironment() throws, confirming the fallback in permitsMethod() will kick in gracefully. All 6 tests pass.
> 
> **4. @DisplayName — Can't find precedent in the codebase.** I went through the existing test classes and none of them use @DisplayName. Adding it here would introduce inconsistency rather than improve it. If there's a specific test class you're referencing that uses it, happy to follow that pattern — but I'd rather stay consistent with what's already there.

Thanks for addressing all the points and for the detailed explanation 
on the profiling test. The reasoning makes sense testGetEnvironmentCalledOnlyOnce 
already guards the correctness.

Vinamra-tech:
Hello @slide just a follow up message on this pr, could the proposed changes mentioned in this pr be reviewed, please?

slide:
I've been inundated with PR's recently (which is a good thing!) so it is taking me a while to get through them all. I have a normal day job that takes priority, but I will make another pass soon on the open PRs and try and get them resolved.

Vinamra-tech:
> I've been inundated with PR's recently (which is a good thing!) so it is taking me a while to get through them all. I have a normal day job that takes priority, but I will make another pass soon on the open PRs and try and get them resolved.

Sure, I should've been more careful with that!! you may review the pr whenever it deems necessary. 
Thank You!!

slide:
Well-engineered optimisation — the `try-finally` guarantee around `beginExecution`/`endExecution` is exactly right, the fallback paths are solid, and the encapsulation of `ENV_CACHE` as private with a package-private accessor is the correct approach. The reviewer feedback was handled well.

### `beginExecution` swallows `InterruptedException` without restoring the interrupt flag

Given the pattern being fixed in #1561 and #1562, it's worth noting that `beginExecution` catches `IOException | InterruptedException` together and doesn't call `Thread.currentThread().interrupt()` for the interrupt case. The fix is the same as those PRs — split the catch block:

```java
} catch (IOException e) {
    LOGGER.log(Level.WARNING, e, () -> "Failed to pre-load environment for " + build.getExternalizableId());
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
    LOGGER.log(Level.WARNING, e, () -> "Failed to pre-load environment for " + build.getExternalizableId());
}
```

### `testBeginExecutionHandlesInterruptedException` doesn't verify the interrupt flag

The test correctly asserts the cache is null after the exception, but doesn't check whether the interrupt flag is restored — which would be the key thing to assert once point 1 above is fixed. Use `Thread.currentThread().isInterrupted()` for the assertion and clear the flag in `tearDown` to avoid test bleed.

### Wildcard import in test file

```java
import static org.mockito.Mockito.*;
```
The codebase uses explicit imports. Please replace with explicit imports.

### `testProfilingGain` uses busy-waiting which may be CI-hostile

The test burns CPU in a tight loop for 5ms × 100 iterations plus warm-up runs — potentially seconds of real CPU time per CI run. Please confirm that the CI configuration filters out `@Tag("performance")` tests, otherwise this will add meaningful overhead to every build.

Vinamra-tech:
@slide I've tried to address each and every point:

**1. InterruptedException — split catch block with interrupt flag restored**
I've splitted the combined `catch (IOException | InterruptedException)` block 
into separate handlers. The `InterruptedException` path now calls 
`Thread.currentThread().interrupt()` before logging, consistent with 
the pattern fixed in #1561 and #1562.

**2. testBeginExecutionHandlesInterruptedException — interrupt flag assertion added**
I've updated the test to assert `Thread.currentThread().isInterrupted()` 
after `beginExecution()` to verify the flag is correctly restored. 
Also, I've added `Thread.interrupted()` in `tearDown()` to clear the flag after 
each test and prevent bleed into subsequent tests.

**3. Wildcard imports — replaced with explicit imports**
I've replaced both `import static org.junit.jupiter.api.Assertions.*` and 
`import static org.mockito.Mockito.*` with explicit imports for only 
what is actually used. The main culprit behind this was actually Intellij 
trying to merge similar import statements into one statement which covers
all the others.

**4. CI overhead from testProfilingGain — excluded at discovery phase**
The hpi plugin pulls in JUnit 5 transitively and intercepts surefire 
configuration, so `-Dgroups` doesn't reach JUnit Platform directly. 
Tag exclusion is handled via `src/test/resources/junit-platform.properties`:

  junit.jupiter.tags.exclude=performance

I've also  added `<excludedGroups>performance</excludedGroups>` to the 
surefire configuration in `pom.xml` as a fallback.

Exclusion happens at the JUnit Platform discovery phase rather than 
runtime, so the test adds zero overhead to CI — it isn't collected 
into the execution plan at all. Confirmed — standard run shows 
`Tests run: 5, Skipped: 0` with `testProfilingGain` absent from 
the plan entirely.

slide:
Thanks!

Vinamra-tech:
Thanks!

PR #1569, opened 2026-04-14
test: add unit tests for EmailExtensionPlugin static initializer

ArpanC6:
### Summary

So there’s a class called **EmailExtensionPlugin** and it has a static initializer block that takes care of setting two system properties **mail.smtp.sendpartial** and **mail.smtps.sendpartial** to **"true"**. But here’s the thing this block runs when the class is loaded and doesn't have any tests to verify its behavior.

### The Issue

The problem is that static initializers run only once when the class is first loaded which means we can’t directly test them. This left two important things untested:

1) Whether the properties are set to **"true"** if they weren’t already defined.

2) Whether the properties stay the same if they were already set to something else.

### What Was Done

To make things easier to test, I extracted the logic from the static initializer block into a separate static method called **initializeSendPartialProperties()**. This allows us to test the behavior without worrying about the class loading. The static block now simply calls this method so the production behavior remains the same.

I also created a new test class **EmailExtensionPluginTest** and added two test cases

1) **whenPropertiesNotSet_shouldSetThemToTrue** This test checks if the properties are set to **"true"** when they aren’t already defined.

2) **whenPropertiesAlreadySet_shouldNotOverrideThem** This one ensures that if the properties already have values they won’t get overwritten.

### Test Setup

To keep everything isolated I made sure to save and clear the properties before each test with **@BeforeEach** and then restore their original values afterward using **@AfterEach**. This ensures no tests interfere with each other and the changes are properly cleaned up.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/EmailExtensionPlugin.java
+++ b/src/main/java/hudson/plugins/emailext/EmailExtensionPlugin.java
@@ -37,6 +37,11 @@ import java.util.Arrays;
  */
 public class EmailExtensionPlugin extends Plugin {
     static {
+        initializeSendPartialProperties();
+    }
+
+    /** Package private for testing only. */
+    static void initializeSendPartialProperties() {
         for (String property : Arrays.asList("mail.smtp.sendpartial", "mail.smtps.sendpartial")) {
             if (System.getProperty(property) == null) {
                 System.setProperty(property, "true");
--- a/src/test/java/hudson/plugins/emailext/EmailExtensionPluginTest.java
+++ b/src/test/java/hudson/plugins/emailext/EmailExtensionPluginTest.java
@@ -1,0 +1,62 @@
+package hudson.plugins.emailext;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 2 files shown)


ArpanC6:
@slide Added unit tests for the static initializer block in EmailExtensionPlugin. 
Happy to make any changes if needed.

slide:
Clean approach — extracting the static initializer logic into a package-private method is the standard pattern for this problem, and the test isolation via `@BeforeEach`/`@AfterEach` including the `null` restoration case is done correctly.

Two minor suggestions:

### Consider a test for partial state

The two tests cover neither property set and both properties already set. A third case — one property set and one not — would confirm the loop handles each property independently and make that behaviour explicit:

```java
@Test
void whenOnlyOnePropertySet_shouldSetTheOther() {
    System.setProperty("mail.smtp.sendpartial", "false");
    EmailExtensionPlugin.initializeSendPartialProperties();
    assertEquals("false", System.getProperty("mail.smtp.sendpartial"));
    assertEquals("true", System.getProperty("mail.smtps.sendpartial"));
}
```

### `"Package private for testing only"` comment

This is a widely accepted pragmatic pattern for static initializers so it's not a blocker, but it's worth being aware that flagging production code as existing purely for test access is generally a mild design smell. An alternative would be to keep the logic in the static block and test it via observable side effects at class load time — though for this specific case the extracted method approach is simpler and fine.

PR #1571, opened 2026-04-21
Replace broad Exception catch with specific exceptions in renderTemplate

ArpanC6:
In the original **renderTemplate** method in EmailExtTemplateAction.java a broad catch (Exception ex) block was used which caught all exceptions without differentiating between various types of errors.

However certain methods, such as evaluate() on **JellyScriptContent** and **ScriptContent** can throw specific exceptions like **MacroEvaluationException** and **IOException.** Additionally calling **stream.toString()** can throw an **UnsupportedEncodingException**. These are distinct failure scenarios and it makes more sense to handle them separately rather than catching them all in one block.

To improve this the broad catch block has been replaced with specific exception handlers for

**1) MacroEvaluationException
2 ) UnsupportedEncodingException
3) IOException
4) InterruptedException**

For **InterruptedException** we now call **Thread.currentThread()**.interrupt() before handling the error. This restores the thread's interrupted status which is the recommended practice when catching and not rethrowing an **InterruptedException.**

### DIFF
--- a/src/main/java/hudson/plugins/emailext/EmailExtTemplateAction.java
+++ b/src/main/java/hudson/plugins/emailext/EmailExtTemplateAction.java
@@ -15,17 +15,17 @@ import hudson.util.StreamTaskListener;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


slide:
This overlaps directly with #1553, please make sure the PR is clean.

ArpanC6:
Closing this in favour of #1553 which covers the same file. Will address the changes there.

PR #1573, opened 2026-04-22
Add test coverage for the class `EmailExtTemplateAction`

akash-manna-sky:
Add test coverage for the class `EmailExtTemplateAction`

### Submitter checklist


akash-manna-sky:
Please review the changes. @slide

slide:
Apologies for the long wait time, but I have been inundated with many pull requests in the lead up to GSoC and I have a day job that takes priority. I will get to them as soon as I can.

slide:
Good set of integration tests — the permission checks cover security-relevant behaviour correctly, the `ACLContext` usage is right, and the `@AfterEach` cleanup with the `IllegalStateException` guard is thoughtful.

### `renderTemplateRendersErrorWhenBuildLookupFails` asserts on `NullPointerException` by name

```java
assertTrue(result[0].contains("NullPointerException"));
```

This ties the test to the current incidental error path. If the code is ever improved to handle a missing build ID gracefully with a meaningful message rather than an NPE, this test will fail even though the behaviour improved. Consider asserting on a stable indicator instead — e.g. that `result[0]` contains `"An error occurred"` and that `result[1]` is empty — rather than pinning on the specific exception class name.

### `doTemplateFileCheckReportsManagedTemplateBehavior` silently bifurcates on environment

The `if/else` on `j.jenkins.getPlugin("config-file-provider") == null` means what this test actually covers depends on which plugins are installed in the CI environment. On some nodes it tests one path, on others a completely different one. These should be split into two separate tests — one that uses `Assumptions.assumeTrue(j.jenkins.getPlugin("config-file-provider") != null)` to skip cleanly when the plugin isn't present, and one that covers the plugin-not-available error independently.

### Classpath template files aren't visible in the diff

`renderTemplateUsesGroovyTemplates` references `"empty-groovy-template-on-classpath.template"` and `renderTemplateUsesJellyTemplates` references `"empty-template-on-classpath.jelly"`. Please confirm these exist in the test resources, or include them as part of this PR if they don't — otherwise these tests will fail with an uninformative error.

akash-manna-sky:
Please review the changes. @slide

akash-manna-sky:
Thanks!

PR #1574, opened 2026-04-23
fix: replace broad Exception catch with specific exceptions in ContentBuilder.transformText()

ArpanC6:
In the transformText method in ContentBuilder.java, there was originally a broad catch block that was catching every exception after the call to TokenMacro.expandAll. The issue with this approach is that it didn't differentiate between different types of errors. Since TokenMacro.expandAll can throw specific exceptions like IOException and InterruptedException it’s more logical to handle them separately based on the type of error that occurred.

So the broad catch block has been replaced with specific handlers

1) IOException: This handles any issues related to input and output operations.

2) InterruptedException: In this case the code now calls Thread.currentThread().interrupt() before logging the error. This is the recommended approach when catching an InterruptedException and not rethrowing it as it ensures that the interrupt flag is properly set.

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
@@ -7,7 +7,6 @@ import hudson.model.BuildListener;
 import hudson.plugins.emailext.ExtendedEmailPublisher;
 import hudson.plugins.emailext.ExtendedEmailPublisherContext;
 import hudson.tasks.Publisher;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


ArpanC6:
@slide all 17 checks are passing. Please let me know if any further changes are needed. Thank you

slide:
The diff shows 0 files changed. Please make sure the PR branch is correct.

ArpanC6:
@slide Fixed the branch now has the correct specific exception handling with the missing IOException import added. The diff should show the changes properly now.

ArpanC6:
@slide The changes from this PR appear to have already been merged into main. Closing this PR as the fix is already in place.

PR #1575, opened 2026-04-23
Add robust `createFor` coverage in `EmailExtTemplateActionFactoryTest` including `null` publishers list handling

akash-manna-sky:
Add robust `createFor` coverage in `EmailExtTemplateActionFactoryTest` including `null` publishers list handling

### Submitter checklist


slide:
The enhancement to `testCreateForReturnsActionWhenEmailExtPresent` is a nice improvement — verifying the type and project reference is more meaningful than just checking the collection size.

### The null publishers list test may not reflect a real scenario

`getPublishersList()` on a real Jenkins project never returns `null` — it always returns an empty `DescribableList`. This test is essentially guarding against a contract violation that won't occur in production. Before merging, it's worth checking whether `createFor()` actually has a null guard for `getPublishersList()` in production code — if it doesn't, this test would expose a real NPE and the fix belongs in `EmailExtTemplateActionFactory`, not just the test. If it does have a null guard, the test is valid but should note that in a comment.

### Unusual null cast pattern

```java
Mockito.doReturn((DescribableList<?, ?>) null).when(project).getPublishersList();
```

Casting `null` to `DescribableList<?, ?>` is unusual. The more idiomatic Mockito form would be:

```java
Mockito.when(project.getPublishersList()).thenReturn(null);
```

### Testing section is blank

Even for a test-only PR, the "Testing done" section should confirm that the tests pass locally — e.g. `mvn test` output or at minimum a note that all tests pass.

PR #1576, opened 2026-04-29
Fix JENKINS-71828: project recipient list now overrides $DEFAULT_RECIPIENTS

### LINKED ISSUE #1407: [JENKINS-71828] Project Recipient List - does not override default list
<p><span class="image-wrap" style=""></span></p>

<p>I used this </p>

<p><span class="image-wrap" style=""></span></p>

<p><span class="image-wrap" style=""></span></p>

<p><font color="#FF0000"><b>but emails sent to my default emails</b></font></p>

---
<details><summary><i>Originally reported by <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=abn_java">abn_java</a>, imported from: <a class="original-jira-link" href="https://issues.jenkins.io/browse/JENKINS-71828" target="_blank">Project Recipient List - does not override default list</a></i></summary>
<i><ul>
<li><b>status</b>: Open
<li><b>priority</b>: Major
<li><b>label(s)</b>: email-ext
<li><b>votes</b>: 0
<li><b>watchers</b>: 1
</ul></i>

<pre>
<p><span class="image-wrap" style=""></span></p>

<p>I used this </p>

<p><span class="image-wrap" style=""></span></p>

<p><span class="image-wrap" style=""></span></p>

<p><font color="#FF0000"><b>but emails sent to my default emails</b></font></p></pre>
</details>
</details>
<ul><li><i>environment</i>: <code>Jenkins 2.401.3<br/>
<br/></code></li></ul>
<details><summary><i>3 attachments</i></summary>

- [image-2023-08-14-18-17-04-106.png](https://raw.githubusercontent.com/jenkinsci/attachments-from-jira-issues-misc/refs/heads/main/attachments/61015/image-2023-08-14-18-17-04-106.png)
  > 
- [image-2023-08-14-18-18-16-368.png](https://raw.githubusercontent.com/jenkinsci/attachments-from-jira-issues-misc/refs/heads/main/attachments/61014/image-2023-08-14-18-18-16-368.png)
  > 
- [image-2023-08-14-18-19-04-677.png](https://raw.githubusercontent.com/jenkinsci/attachments-from-jira-issues-last/refs/heads/main/attachments/61013/image-2023-08-14-18-19-04-677.png)
  > 
</details>

124padmini:
## Problem
When a project sets its own recipient list explicitly (without $DEFAULT_RECIPIENTS), 
the $DEFAULT_RECIPIENTS token in the trigger-level list was still being expanded to 
the global default recipients, causing both lists to be merged instead of the project 
list taking precedence.

## Fix
In `ContentBuilder.transformText()`, detect when the project has explicitly set its 
own recipient list (non-blank and not delegating to $DEFAULT_RECIPIENTS). In that case, 
use the project recipient list as the replacement for $DEFAULT_RECIPIENTS instead of 
the global default, so the project-level setting correctly overrides the global one.

## Testing
- Existing `ContentBuilderTest` tests pass with no changes
- Spotless formatting check passes

Fixes #1407

### DIFF
--- a/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
+++ b/src/main/java/hudson/plugins/emailext/plugins/ContentBuilder.java
@@ -61,8 +61,17 @@ public final class ContentBuilder {
                 noNull(context.getPublisher().getDescriptor().getDefaultBody()));
         String defaultExtSubject = Matcher.quoteReplacement(
                 noNull(context.getPublisher().getDescriptor().getDefaultSubject()));
+        // If the project has explicitly set its own recipient list (not delegating
+        // to $DEFAULT_RECIPIENTS), honour that override by not expanding
+        // $DEFAULT_RECIPIENTS in the trigger-level list either (fixes JENKINS-71828).
+        String projectRecipientList = noNull(context.getPublisher().recipientList);
+        boolean projectOverridesRecipients = !projectRecipientList.isEmpty()
+                && !projectRecipientList.contains("$DEFAULT_RECIPIENTS")
+                && !projectRecipientList.contains("${DEFAULT_RECIPIENTS}");
         String defaultRecipients = Matcher.quoteReplacement(
-                noNull(context.getPublisher().getDescriptor().getDefaultRecipients()));
+                projectOverridesRecipients
+                        ? projectRecipientList
+                        : noNull(context.getPublisher().getDescriptor().getDefaultRecipients()));
         String defaultExtReplyTo = Matcher.quoteReplacement(
                 noNull(context.getPublisher().getDescriptor().getDefaultReplyTo()));
         String defaultPresendScript = Matcher.quoteReplacement(


124padmini:
@slide this PR fixes JENKINS-71828 — all 17 CI checks passing, Spotless clean, no conflicts. Would appreciate a review when you get a chance.

slide:
The intent here is reasonable — respecting a project-level recipient list over the global default is a legitimate feature — but the implementation has some issues that need to be worked through before this is mergeable.

### The logic may not be doing what's intended

The fix replaces the value that `$DEFAULT_RECIPIENTS` expands to with the project recipient list. But `$DEFAULT_RECIPIENTS` is a token that appears in the *trigger-level* recipient field, not in `projectRecipientList`. What this change actually does is: when a trigger uses `$DEFAULT_RECIPIENTS`, it now resolves to the project recipient list instead of the global default. That may be the desired behaviour for JENKINS-71828, but it's non-obvious and needs a clearer explanation of the intended before/after semantics to validate.

### The string check for `$DEFAULT_RECIPIENTS` is fragile

```java
!projectRecipientList.contains("$DEFAULT_RECIPIENTS")
&& !projectRecipientList.contains("${DEFAULT_RECIPIENTS}")
```

This is a plain string match on the project recipient list, checking for a token that conceptually belongs in the trigger recipient field. It will silently fail for whitespace variations or tokens embedded in larger expressions, and it's unclear why the project recipient list would contain `$DEFAULT_RECIPIENTS` in the first place.

### No tests added

This is a behaviour change to the core recipient resolution path. There are no tests verifying the fix works as described, or that existing behaviour is preserved when `$DEFAULT_RECIPIENTS` is explicitly included in the project recipient list or left at the default.

### JENKINS-71828 isn't explained in the PR description

The issue number appears in the title but the description doesn't explain the expected before/after behaviour. Please add enough context that reviewers can validate whether the fix is actually correct without having to look up the Jira issue.

PR #1577, opened 2026-04-29
Fix JENKINS-68518: retry email send on transient SMTP 4xx errors

### LINKED ISSUE #1380: [JENKINS-68518] Email-ext treats transient SMTP errors as fatal errors
<p>I have an email server that occasionally returns an SMTP 421 error when Jenkins tries to send an email.  According to the SMTP specification, this is a transient error, however email-ext treats it as a permanent error and gives up on sending the message.</p>

<p>It would be better if email-ext could recognize SMTP 4yz errors as transient, and wait and try again instead of failing immediately.</p>

---
<details><summary><i>Originally reported by  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=anfrind">anfrind</a>, imported from: <a class="original-jira-link" href="https://issues.jenkins.io/browse/JENKINS-68518" target="_blank">Email-ext treats transient SMTP errors as fatal errors</a></i></summary>
<i><ul>
<li><b>assignee</b>:  <a href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=slide_o_mix">slide_o_mix</a>
<li><b>status</b>: Open
<li><b>priority</b>: Minor
<li><b>label(s)</b>: email-ext, smtp
<li><b>votes</b>: 0
<li><b>watchers</b>: 1
</ul></i>

<pre>
<p>I have an email server that occasionally returns an SMTP 421 error when Jenkins tries to send an email.  According to the SMTP specification, this is a transient error, however email-ext treats it as a permanent error and gives up on sending the message.</p>

<p>It would be better if email-ext could recognize SMTP 4yz errors as transient, and wait and try again instead of failing immediately.</p></pre>
</details>
</details>

124padmini:
## Problem
Transient SMTP errors (4xx response codes like 421, 450, 451, 452) were 
treated as fatal, causing email sends to fail permanently when the SMTP 
server was temporarily unavailable or rate-limiting.

Only `SocketException` and `ConnectException` triggered a retry. All 
other `MessagingException` types broke immediately.

## Fix
Added `isTransientSmtpError()` helper that detects SMTP 4xx response 
codes in the exception message. Both the `SendFailedException` and 
`MessagingException` catch blocks now retry on transient errors, 
consistent with existing socket error retry behavior.

## Testing
- All existing `ExtendedEmailPublisherTest` tests pass
- Spotless formatting check passes

Fixes #1380

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisher.java
@@ -527,6 +527,23 @@ public class ExtendedEmailPublisher extends Notifier {
         return true;
     }
 
+    /**
+     * Returns true if the SMTP error is transient (4xx response code) and the
+     * send should be retried (fixes JENKINS-68518).
+     */
+    private static boolean isTransientSmtpError(Exception e) {
+        String msg = e.getMessage();
+        if (msg == null) {
+            return false;
+        }
+        // SMTP 4xx codes indicate transient server-side failures
+        return msg.matches("^4\\d\\d.*")
+                || msg.contains("421")
+                || msg.contains("450")
+                || msg.contains("451")
+                || msg.contains("452");
+    }
+
     @SuppressFBWarnings("REC_CATCH_EXCEPTION")
     boolean sendMail(ExtendedEmailPublisherContext context) {
         try {
@@ -620,6 +637,13 @@ public class ExtendedEmailPublisher extends Notifier {
                                         .println("Socket error sending email, retrying once more in 10 seconds...");
                                 transport.close();
                                 Thread.sleep(10000);
+                            } else if (isTransientSmtpError(e)) {
+                                context.getListener()
+                                        .getLogger()
+                                        .println(
+                                                "Transient SMTP error sending email, retrying once more in 10 seconds...");
+                                transport.close();
+                                Thread.sleep(10000);
                             } else {
                                 Address[] addresses = e.getValidSentAddresses();
                                 if (addresses != null && addresses.length > 0) {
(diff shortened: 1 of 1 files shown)


124padmini:
@slide this PR fixes JENKINS-68518 — adds retry logic for transient SMTP 4xx errors, consistent with existing socket error retry behavior. All 17 CI checks are passing. Would appreciate a review.

slide:
The structure is right — extracting the check into `isTransientSmtpError()` makes it testable and reusable, and applying it in both catch blocks is thorough. A few issues with the implementation though:

### The regex and `contains` checks are redundant and inconsistent

`msg.matches("^4\\d\\d.*")` already matches any string starting with a 4xx code — including 421, 450, 451, and 452. The four subsequent `msg.contains(...)` calls are completely redundant when the message starts with the code. Worse, `msg.contains("421")` will also match messages that merely *mention* 421 anywhere in the string — e.g. `"expected 250 but got 421 from relay"` — which could cause false positives. Pick one consistent approach and drop the other:

```java
// regex only — cleaner and unambiguous
return msg.matches("^4\\d\\d.*");
```

### String parsing is fragile — consider using `SMTPSendFailedException.getReturnCode()` instead

SMTP error messages aren't guaranteed to start with the response code — this depends on the JavaMail implementation and the SMTP server. A more robust approach would be to walk the exception's cause chain looking for `SMTPSendFailedException`, which exposes `getReturnCode()` directly:

```java
Throwable cause = e;
while (cause != null) {
    if (cause instanceof SMTPSendFailedException smtpEx) {
        int code = smtpEx.getReturnCode();
        return code >= 400 && code < 500;
    }
    cause = cause.getCause();
}
return false;
```

### No tests added

`isTransientSmtpError()` is a pure static method that's straightforward to unit test. Given the redundancy and fragility issues above, tests that validate the matching logic and guard against regressions would be valuable before this merges.

### Document the precedence between `isSocketConnectionException` and `isTransientSmtpError`

If both conditions could match the same exception, only the socket path runs due to ordering. This isn't documented anywhere and the dependency is fragile. A short comment explaining the precedence would help future maintainers.

PR #1580, opened 2026-05-21
Stream build log attachment instead of buffering entire content in memory

RanjitCS:
`LogFileDataSource.getInputStream()` buffers the entire console log into a `ByteArrayOutputStream` before returning it, causing `OutOfMemoryError` on large logs regardless of `maxAttachmentSize` configuration. 

### Testing done
Added unit tests which OOM without the fix.
Tested in a jenkins instance which shows no OOM after the fix

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
+++ b/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
@@ -111,7 +111,9 @@ public class AttachmentUtils implements Serializable {
                         try (out) {
                             long logFileLength = run.getLogText().length();
                             long pos = 0;
-                            pos = run.getLogText().writeLogTo(pos, out);
+                            while (pos < logFileLength) {
+                                pos = run.getLogText().writeLogTo(pos, out);
+                            }
                         } catch (IOException e) {
                             LOGGER.log(Level.FINE, "Error streaming build log", e);
                         }
--- a/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
+++ b/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
@@ -115,10 +115,10 @@ public class AttachmentUtils implements Serializable {
                                 pos = run.getLogText().writeLogTo(pos, out);
                             }
                         } catch (IOException e) {
-                            LOGGER.log(Level.FINE, "Error streaming build log", e);
+                            LOGGER.log(Level.WARNING, "Error streaming build log", e);
                         }
                     },
-                    "email-ext-log-writer-" + run.getFullDisplayName());
+                    "email-ext-log-writer-" + run.getNumber());
             writer.setDaemon(true);
             writer.start();
 
--- a/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
+++ b/src/main/java/hudson/plugins/emailext/AttachmentUtils.java
@@ -18,15 +18,17 @@ import jakarta.mail.Multipart;
 import jakarta.mail.Part;
 import jakarta.mail.internet.MimeBodyPart;
 import jakarta.mail.internet.MimeUtility;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
 import java.io.Serial;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.apache.commons.lang3.StringUtils;
 import org.jenkinsci.plugins.variant.OptionalExtension;
 
@@ -35,6 +37,8 @@ import org.jenkinsci.plugins.variant.OptionalExtension;
  *
  */
 public class AttachmentUtils implements Serializable {
+
+    private static final Logger LOGGER = Logger.getLogger(AttachmentUtils.class.getName());
 
     @Serial
     private static final long serialVersionUID = 1L;
@@ -99,17 +103,24 @@ public class AttachmentUtils implements Serializable {
 
         @Override
         public InputStream getInputStream() throws IOException {
-            InputStream res;
-            long logFileLength = run.getLogText().length();
-            long pos = 0;
-            ByteArrayOutputStream bao = new ByteArrayOutputStream();
+            PipedInputStream in = new PipedInputStream(65536);
+            PipedOutputStream out = new PipedOutputStream(in);
 
-            while (pos < logFileLength) {
-                pos = run.getLogText().writeLogTo(pos, bao);
-            }
+            Thread writer = new Thread(
+                    () -> {
+                        try (out) {
+                            long logFileLength = run.getLogText().length();
+                            long pos = 0;
+                                pos = run.getLogText().writeLogTo(pos, out);
+                        } catch (IOException e) {
+                            LOGGER.log(Level.FINE, "Error streaming build log", e);
+                        }
+                    },
+                    "email-ext-log-writer-" + run.getFullDisplayName());
+            writer.setDaemon(true);
+            writer.start();
 
-            res = new ByteArrayInputStream(bao.toByteArray());
-            return res;
+            return in;
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/emailext/AttachmentUtils.java
```diff
@@ -99,17 +103,24 @@ public LogFileDataSource(Run<?, ?> run) {
...
-            return res;
+            PipedInputStream in = new PipedInputStream(65536);
+            PipedOutputStream out = new PipedOutputStream(in);
+
+            Thread writer = new Thread(
+                    () -> {
+                        try (out) {
+                            long logFileLength = run.getLogText().length();
+                            long pos = 0;
+                            pos = run.getLogText().writeLogTo(pos, out);
```
nevingeorgesunny: I dont know  about PipedInputStream

but shouldnt we put `pos = run.getLogText().writeLogTo(pos, out);` in a loop like 

```
  while (pos < logFileLength) {  // Keep going until we've written everythin
      pos = run.getLogText().writeLogTo(pos, out);
  }
```

RanjitCS: yeah, you are right i had the while loop but i think i reset it while testing, will fix, thanks

RanjitCS: https://github.com/jenkinsci/email-ext-plugin/pull/1580/commits/773709819e2d10d8119d538d22b72c6af08abefe

nevingeorgesunny (approved):
LGTM

src/main/java/hudson/plugins/emailext/AttachmentUtils.java
```diff
@@ -99,17 +103,26 @@ public LogFileDataSource(Run<?, ?> run) {
...
+            PipedInputStream in = new PipedInputStream(65536);
+            PipedOutputStream out = new PipedOutputStream(in);
+
+            Thread writer = new Thread(
+                    () -> {
+                        try (out) {
+                            long logFileLength = run.getLogText().length();
+                            long pos = 0;
+                            while (pos < logFileLength) {
+                                pos = run.getLogText().writeLogTo(pos, out);
```
jglick, line 115: Note that this strips annotations. Otherwise you could I guess use `Run.getLogInputStream` though first https://github.com/jenkinsci/workflow-job-plugin/blob/e4bd0dbfcbbb7f21d942fe84c817811c51100758/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L1151 would need to be addressed.

RanjitCS: AFAICT, this was already pre-existing behavior `AttachmentUtils.LogFileDataSource.getInputStream()` already called `run.getLogText().writeLogTo(pos, bao)`, which then calls `AnnotatedLargeText.writeLogTo(long, OutputStream)` and strips annotations via `PlainTextConsoleOutputStream`.

Switching to `Run.getLogInputStream()` would re-introduce the OOM for pipeline builds as https://github.com/jenkinsci/workflow-job-plugin/blob/e4bd0dbfcbbb7f21d942fe84c817811c51100758/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L1152 buffers the full log into a ByteArrayOutputStream, which is the same thing we are trying to avoid here

jglick: Exactly, just noting for context.

slide:
Good find! 

### IOException in the writer thread is silently swallowed

If `writeLogTo` fails partway through, the writer thread catches the exception, logs it at `FINE` (essentially invisible in production), closes the `PipedOutputStream`, and exits. The reader will see EOF with no indication the log was truncated — the email attachment will be silently incomplete. This should be logged at `WARNING` at minimum:

```java
} catch (IOException e) {
    LOGGER.log(Level.WARNING, "Build log streaming interrupted — attachment may be truncated", e);
}
```

### Thread name may contain special characters

`run.getFullDisplayName()` can contain slashes, hashes, and brackets from build display names, which can cause issues in some JVM thread dump tools. Consider using the build number instead:

```java
"email-ext-log-writer-" + run.getNumber()
```

### The test generates 50MB of output on every CI run

This is a heavy test that will meaningfully slow down CI. Consider adding a `@Tag("slow")` annotation so it can be excluded from fast CI runs. It's also worth confirming the test JVM has a heap limit set low enough to actually reproduce the OOM without the fix — otherwise the test doesn't prove what it claims.

RanjitCS:
@slide, thanks for the review i addressed the first two comments, but for the third one 

>Consider adding a @Tag("slow")

this will not actually do anything unless the surefire config is updated with `excludedGroups`, do you want me to setup this?

>It's also worth confirming the test JVM has a heap limit

this would also require a seperate profile that only applies the heap limit to this test, for now i have added a doc explaining the limit required. let me know if you want me to set up the profile

jglick:
> issues in some JVM thread dump tools

Really? Jenkins has all sorts of thread names for different purposes. I have never heard of such a problem.

src/test/java/hudson/plugins/emailext/AttachmentUtilsTest.java
```diff
@@ -150,6 +150,8 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
                 containsString("build.log")); // zips have plain text filename in them
     }
 
+    // To verify this test actually catches the regression, run with -Xmx128m
```
jglick: If you want to enforce a particular JVM setting for a particular test, you must use `RealJenkinsRule`.

src/main/java/hudson/plugins/emailext/AttachmentUtils.java
```diff
@@ -99,17 +103,26 @@ public LogFileDataSource(Run<?, ?> run) {
...
+                            }
+                        } catch (IOException e) {
+                            LOGGER.log(Level.WARNING, "Error streaming build log", e);
```
jtnord: I have a feeling that when the reader has read enough of it's lines it will close its stream. That should cause the blocked writer to fail when it attempts to write as the pipe is broke.
Won't this then log a warningwhenever the log is bigger than the max buffer?

src/test/java/hudson/plugins/emailext/AttachmentUtilsTest.java
RanjitCS: https://github.com/jenkinsci/email-ext-plugin/pull/1580/commits/4368fbab8685d07c5bcc5be1611b6f118a199436

src/main/java/hudson/plugins/emailext/AttachmentUtils.java
RanjitCS: ah you are right, i am going to revert the warning to fine. When the reader finishes and closes the stream, the writer gets a broken-pipe IOException which is expected and not worth surfacing as a warning. Instead of trying to find the specific exception that will occur and logging that as fine, its better imo to just keep the entire log was WARN. If `writeLogTo` itself fails for a genuine reason, the reader already sees `IOException("Write end dead")` from PipedInputStream and that propagates to the callers error handling so we should be safe.

RanjitCS:
@slide could you please review/merge whenever you get some time?

slide (approved):

RanjitCS:
@slide, thanks for the merge, would it be possible to release as well?

jglick:
I guess this was missing a label for CD. https://github.com/jenkins-infra/interesting-category-action/issues/1

jglick:
The changelog for https://github.com/jenkinsci/email-ext-plugin/releases/tag/2038.v7b_8817a_499d9 is nearly empty despite containing dozens of changes. It seems that e.g. https://github.com/jenkinsci/email-ext-plugin/pull/1579 lacked a label (`bug` would have been appropriate in that case) and so not only did not trigger a release as it should have, but is omitted from the release notes when one _was_ created.

PR #1581, opened 2026-05-25
Add OAuth2 credential selection to MailAccount dropdown

Faheem12005:
- This PR adds Oauth Credentials plugin as a dependency and allows Oauth Credentials to be selected from the dropdown in the config page.

  It handles this by checking if Oauth2 is enabled, and deciding to display credentials of type:
  - `StandardUsernamePasswordCredentials`
  - `StandardUsernameOAuth2Credentials`

- I decided to go with `StandardUsernameOAuth2Credentials` since both plugins that currently work using Oauth Credentials, being entra-oauth and google-oauth implement using it.

- This PR also adds tests to check if credentials are properly filtered and validated based on whether OAuth2 is enabled.

Note: I also would have liked to test with entra-oauth but I was unable to find it on the available plugins page, I felt it was sufficient to test with google-oauth since both implement the same credential type so should functionally work the same for these changes
### Testing done

- Added unit tests verifying credential dropdown filtering and validation.

- I've also manually tested by installing google-oauth plugin on a dev jenkins instance and setting up a dummy credential and verified that credentials are properly filtered and revalidated when toggling Oauth2.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -47,8 +47,6 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
     private String advProperties;
     private boolean defaultAccount;
 
-    private boolean useOAuth2;
-
     @Deprecated
     public MailAccount(JSONObject jo) {
         address = Util.nullify(jo.optString("address", null));
@@ -114,12 +112,9 @@ public class MailAccount extends AbstractDescribableImpl<MailAccount> {
         }
 
         @SuppressWarnings({"lgtm[jenkins/csrf]", "unused"}) // Used by stapler
-        public ListBoxModel doFillCredentialsIdItems(
-                @AncestorInPath Item item, @QueryParameter String credentialsId, @QueryParameter boolean useOAuth2) {
+        public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item item, @QueryParameter String credentialsId) {
 
             final StandardListBoxModel result = new StandardListBoxModel();
-            Class<? extends StandardUsernameCredentials> credentialType =
-                    useOAuth2 ? StandardUsernameOAuth2Credentials.class : StandardUsernamePasswordCredentials.class;
             if (item == null) {
                 if (!Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
                     return result.includeCurrentValue(credentialsId);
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/plugins/emailext/MailAccount.java
```diff
@@ -112,9 +114,12 @@ public String getDisplayName() {
         }
 
         @SuppressWarnings({"lgtm[jenkins/csrf]", "unused"}) // Used by stapler
-        public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item item, @QueryParameter String credentialsId) {
+        public ListBoxModel doFillCredentialsIdItems(
+                @AncestorInPath Item item, @QueryParameter String credentialsId, @QueryParameter boolean useOAuth2) {
```
slide: I think instead of adding an other parameter here, we can just use one of these to match both credential types below:

https://javadoc.jenkins.io/plugin/credentials/com/cloudbees/plugins/credentials/CredentialsMatchers.html#either(com.cloudbees.plugins.credentials.CredentialsMatcher,com.cloudbees.plugins.credentials.CredentialsMatcher)

or 

https://javadoc.jenkins.io/plugin/credentials/com/cloudbees/plugins/credentials/CredentialsMatchers.html#anyOf(com.cloudbees.plugins.credentials.CredentialsMatcher...)

src/main/java/hudson/plugins/emailext/MailAccount.java
```diff
@@ -139,7 +144,10 @@ public FormValidation doCheckCredentialsId(
                 @AncestorInPath Item item,
                 @QueryParameter String value,
                 @QueryParameter boolean useSsl,
-                @QueryParameter boolean useTls) {
+                @QueryParameter boolean useTls,
+                @QueryParameter boolean useOAuth2) {
```
slide: Same comment as above about using a different matcher.

slide (changes requested):
Overall pretty good, just a couple of comments (though they are pretty much the same thing).

src/main/java/hudson/plugins/emailext/MailAccount.java
Faheem12005: Oh okay. I'd gone with this approach as there was an existing parameter useOAuth2. are you suggesting that rather than having the user explicitly select whether to use OAuth2, we let the credential decide that? If so I could remove the useOauth2 checkbox from the UI as well

### Edit: 
We'd still need the checkbox to set the XOAUTH2 property in `ExtendedEmailPublisherDescriptor.java`. The reason I've gone with adding useOauth2 as a parameter is so that when Oauth2 is enabled, only Oauth Credentials are visible to select from the dropdown and vice versa. If i was to use the credential matcher alone I wouldn't be able to do this I believe.

slide: I think it would be good to just determine that OAuth 2.0 should be used if the credential is an OAuth 2.0 credential. The people setting up the credentials and such will be admins, so they will know which credentials are which.

Faheem12005: I agree! I've made some changes and removed the useOauth2 checkbox as well. this does mean we'd need to resolve the credential type in `ExtendedEmailPublisherDescriptor.java` to set the auth mechansim, so I've left that as a todo.

PR #1583, opened 2026-05-30
add Test Oauth2 Credentials class for unit/integration tests

Faheem12005:
Adds a test only implementation of an Oauth2 credential that Jenkins tests can use to verify behavior.  The class mirrors the structure of existing provider implementations, and has a test OAuth2ScopeRequirement in case scope behavior might need to be tested.
This class can be used for credential discovery and lookup, exercising code paths that request an access token, and validating OAuth UI, etc

Added a small integration test to verify the test credential can be properly registered through the credential API since I'm planning on reusing it for subsequent OAuth related tests

### Submitter checklist


Faheem12005:
I'm pretty sure tests are failing as oauth-credentials is not part of the dependencies of the main branch yet

Faheem12005:
I’ve temporarily rebased this PR onto my earlier PR that adds the oauth-credentials dependency so the CI can run. Once the dependency PR is merged, I’ll rebase this back onto main

slide:
Looks like there is a conflict now, can you take a look?

Faheem12005:
> Looks like there is a conflict now, can you take a look?

Yes I had this rebased on top of another PR that hadn't been merged, so it's commits were included in this branch as well. I've now rebased this onto gsoc2026, and should only contain relevant changes

PR #1589, opened 2026-06-17
Add single attempt retry logic to authenticator

Faheem12005:
Modified the authenticatorProvider to retry once when there is failure during OAuth token fetching. This should handle transient errors such as if the OAuth provider was down or other network errors that may be encountered.

### Testing done
Manually tested both the OAuth and normal SMTP authentication code paths. Both still work as expected.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
+++ b/src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
@@ -224,52 +224,65 @@ public final class ExtendedEmailPublisherDescriptor extends BuildStepDescriptor<
                     }
 
                     String address = acc.isDefaultAccount() ? getAdminAddress() : acc.getAddress();
+                    final int TOKEN_ATTEMPTS = 2;
+                    final long RETRY_DELAY_MS = 2000;
 
                     if (c instanceof GoogleRobotPrivateKeyCredentials googleCred) {
                         GoogleCredential credential = googleCred
                                 .getGoogleCredential(new MailScopeRequirement())
                                 .createDelegated(address);
 
-                        try {
-                            credential.refreshToken();
-                        } catch (IOException e) {
-                            LOGGER.log(Level.WARNING, "Failed to obtain access token, retrying once.", e);
-
-                            try {
-                                Thread.sleep(2000);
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
```diff
@@ -223,25 +223,57 @@ protected PasswordAuthentication getPasswordAuthentication() {
...
 
+                    String address = acc.isDefaultAccount() ? getAdminAddress() : acc.getAddress();
+
                     if (c instanceof GoogleRobotPrivateKeyCredentials googleCred) {
                         GoogleCredential credential = googleCred
                                 .getGoogleCredential(new MailScopeRequirement())
-                                .createDelegated(acc.isDefaultAccount() ? getAdminAddress() : acc.getAddress());
+                                .createDelegated(address);
+
                         try {
```
slide: Can we perhaps make this into a loop so that in the future, if desired, the number of retries can be configured?

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
```diff
@@ -223,25 +223,57 @@ protected PasswordAuthentication getPasswordAuthentication() {
...
-                                acc.isDefaultAccount() ? getAdminAddress() : acc.getAddress(), token);
+
+                        return new PasswordAuthentication(address, token);
                     }
 
                     if (c instanceof EntraOAuthCredentials entraCred) {
-                        return new PasswordAuthentication(
-                                acc.isDefaultAccount() ? getAdminAddress() : acc.getAddress(),
-                                Secret.toString(entraCred.getAccessToken(null)));
+                        Secret tokenSecret = entraCred.getAccessToken(null);
```
slide: Same loop comment here

src/main/java/hudson/plugins/emailext/ExtendedEmailPublisherDescriptor.java
Faheem12005: sounds good. I've wrapped the retry logic in for-loops for both and extracted the wait time and no of attempts into constants at the top of the function

PR #1591, opened 2026-06-18
Add test coverage of `EmailExtWatchAction` class 

akash-manna-sky:
Add test coverage of `EmailExtWatchAction` class 
### Submitter checklist

### DIFF
--- a/src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
+++ b/src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
@@ -64,6 +64,25 @@ class EmailExtWatchActionTest {
                 .to("alice"));
     }
 
+    /**
+     * Helper method to set up the ExtendedEmailPublisherDescriptor with watching enabled/disabled
+     * and register the AlwaysTrigger descriptor. Eliminates boilerplate repeated across tests.
+     */
+    private void setupWatchingEnabled(boolean enabled) {
+        ExtendedEmailPublisherDescriptor descriptor =
+                j.jenkins.getDescriptorByType(ExtendedEmailPublisherDescriptor.class);
+        if (descriptor == null) {
+            descriptor = new ExtendedEmailPublisherDescriptor();
+            j.jenkins.getExtensionList(Descriptor.class).add(0, descriptor);
+        }
+        descriptor.setWatchingEnabled(enabled);
+
+        if (enabled) {
+            AlwaysTrigger.DescriptorImpl alwaysDesc = new AlwaysTrigger.DescriptorImpl();
+            j.jenkins.getExtensionList(Descriptor.class).add(0, alwaysDesc);
+        }
+    }
+
     @Test
     void testBasicProperties() {
         assertNull(action.getIconFileName());
@@ -124,7 +143,7 @@ class EmailExtWatchActionTest {
     }
 
     @Test
-    void testUserSpecificMethods() throws Exception {
+    void testStartStopWatching() throws Exception {
         User user = User.getById("alice", true);
 
         try (ACLContext ctx = ACL.as(user)) {
@@ -142,7 +161,14 @@ class EmailExtWatchActionTest {
 
             action.stopWatching();
             assertFalse(action.isWatching(user));
+        }
+    }
 
+    @Test
+    void testGetTriggersIgnoresWrongProject() throws Exception {
+        User user = User.getById("alice", true);
+
+        try (ACLContext ctx = ACL.as(user)) {
             List<EmailTrigger> triggers = new ArrayList<>();
             triggers.add(new AlwaysTrigger(Collections.emptyList(), "", "", "", "", "", 0, ""));
             EmailExtWatchAction.UserProperty userPropWrongProject =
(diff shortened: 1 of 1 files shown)


src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
```diff
@@ -0,0 +1,476 @@
...
+            assertFalse(action.isWatching());
+
+            action.startWatching();
+            action.stopWatching();
+        }
+    }
+
+    @Test
+    void testDoStopWatching() throws Exception {
+        ExtendedEmailPublisherDescriptor descriptor =
```
slide: Maybe convert this duplicated code into a helper method like `setupWatchingEnabled(boolean enabled)`

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
```diff
@@ -0,0 +1,476 @@
...
+            assertTrue(action.isWatching(user));
+            assertFalse(userProp.getTriggers().isEmpty());
+
+            StaplerRequest2 req = mock(StaplerRequest2.class);
+            StaplerResponse2 rsp = mock(StaplerResponse2.class);
+
+            action.doStopWatching(req, rsp);
+
+            assertFalse(action.isWatching(user));
+            assertTrue(userProp.getTriggers().isEmpty());
```
slide, line 256: Should probably fetch the property again so you are not using the local object, but rather the one that Jenkins is managing.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
```diff
@@ -0,0 +1,476 @@
...
+            descriptor = new ExtendedEmailPublisherDescriptor();
+            j.jenkins.getExtensionList(Descriptor.class).add(0, descriptor);
+        }
+        descriptor.setWatchingEnabled(false);
+
+        User user = User.getById("alice", true);
+        try (ACLContext ctx = ACL.as(user)) {
+            List<EmailTrigger> triggers = new ArrayList<>();
+            triggers.add(new AlwaysTrigger(Collections.emptyList(), "", "", "", "", "", 0, ""));
+            EmailExtWatchAction.UserProperty userProp =
```
slide, line 303: This test may be better if you call `action.doStartWatching()` instead of setting things up manually. That would follow the full flow and give more confidence if the test passes.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
```diff
@@ -0,0 +1,476 @@
...
+        User user = User.getById("alice", true);
+        try (ACLContext ctx = ACL.as(user)) {
+            StaplerRequest2 req = mock(StaplerRequest2.class);
+            StaplerResponse2 rsp = mock(StaplerResponse2.class);
+
+            action.doConfigSubmit(req, rsp);
+
+            assertFalse(action.isWatching(user));
+            EmailExtWatchAction.UserProperty userProp = user.getProperty(EmailExtWatchAction.UserProperty.class);
+            if (userProp != null) {
```
slide: If userProp is null at this point, it means something is wrong, so it should fail the test. It would be best to turn this null check into an assertNotNull on userProp.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
```diff
@@ -0,0 +1,476 @@
...
+        when(req.bindJSONToList(EmailTrigger.class, json)).thenReturn(triggers);
+        EmailExtWatchAction.UserProperty prop2 = descriptor.newInstance(req, json);
+        assertEquals(1, prop2.getTriggers().size());
+
+        EmailExtWatchAction.UserProperty prop3 = descriptor.newInstance((StaplerRequest2) null, json);
+        assertTrue(prop3.getTriggers().isEmpty());
+    }
+
+    @Test
+    void testUserSpecificMethods() throws Exception {
```
slide: Might be good to split this test into multiple tests, like maybe `testStartStopWatching`, `testGetTriggersIgnoresWrongProject`, `testDisplayNameChangesWhenWatching`, etc. This will make it easier to diagnose a test failure.

slide:
Please add a comment to items that you have addressed in my review. This makes it easier to verify that you have completed them.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
akash-manna-sky: All repeated 5-line descriptor setup blocks replaced with a single `setupWatchingEnabled(true/false)` call, registering `AlwaysTrigger.DescriptorImpl` only when enabled.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
akash-manna-sky: Instead of manually constructing and adding a `UserProperty`, the test now calls `doConfigSubmit` first to set up the watching state through the full production code path.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
akash-manna-sky: In `testDoConfigSubmit_Disabled`, correctly uses `assertNull(userProp)` since no property should be created when watching is disabled.

src/test/java/hudson/plugins/emailext/watching/EmailExtWatchActionTest.java
akash-manna-sky: Split into `testStartStopWatching`, `testGetTriggersIgnoresWrongProject`, and `testDisplayNameChangesWhenWatching` for easier failure diagnosis.

slide (approved):

akash-manna-sky:
Thanks!

PR #1596, opened 2026-07-06
Add test coverage for `UserProvidedContentInputStream()` class

akash-manna-sky:
Add test coverage for `UserProvidedContentInputStream()` class
### Submitter checklist

### DIFF
--- a/src/test/java/hudson/plugins/emailext/plugins/content/UserProvidedContentInputStreamTest.java
+++ b/src/test/java/hudson/plugins/emailext/plugins/content/UserProvidedContentInputStreamTest.java
@@ -1,0 +1,368 @@
+package hudson.plugins.emailext.plugins.content;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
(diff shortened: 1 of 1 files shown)


akash-manna-sky:
Hi @slide, could you please review this PR?

slide (changes requested):
Thanks for putting this together. Adding coverage here is useful, and I appreciate the effort to exercise the wrapper more thoroughly.

Most of these tests seem reasonable for a simple delegating InputStream, especially the  read/skip/available/reset exception paths. I do have one small suggestion, though: a couple of the assertions feel like they may be testing implementation quirks more than behavior we really want to preserve long-term.

In particular: the equals test currently expects assertNotEquals(s, s), which makes equality non-reflexive and may be a bit surprising to lock in as intended behavior the close test verifies two close() calls because the stream is closed explicitly and then again by try-with-resources, so it may be stronger to check the delegation without depending on double-close

Would you mind tightening those a bit? With those adjusted, this looks like a nice incremental coverage improvement.

PR #1600, opened 2026-08-15
Add test email feature

Faheem12005:
This PR adds a test-email button to the default mail account section on the system configuration page, so admins can verify their setup without having to save config and trigger a real build.

### Testing done
Tested with Gmail using app-passwords for the username password case, wasn't able to test for OAuth2 as no environment available (trial ran out) however as we are still reaching the google endpoint, we can assume it should work as expected

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/emailext/MailAccount.java
+++ b/src/main/java/hudson/plugins/emailext/MailAccount.java
@@ -24,8 +24,19 @@ import hudson.util.FormValidation;
 import hudson.util.FormValidation.Kind;
 import hudson.util.ListBoxModel;
 import hudson.util.Secret;
+import jakarta.mail.AuthenticationFailedException;
+import jakarta.mail.Message;
+import jakarta.mail.MessagingException;
+import jakarta.mail.Session;
+import jakarta.mail.Transport;
+import jakarta.mail.internet.AddressException;
+import jakarta.mail.internet.InternetAddress;
+import jakarta.mail.internet.MimeMessage;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import jenkins.security.FIPS140;
 import net.sf.json.JSONObject;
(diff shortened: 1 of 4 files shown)


src/main/resources/hudson/plugins/emailext/MailAccount/config.groovy
```diff
@@ -33,3 +33,17 @@ f.advanced {
         f.textarea()
     }
 }
+
+if(instance?.defaultAccount) {
```
slide: Would we want to be able to test with any account or just the default?

slide (commented):
In general looks very nice. Just one small question.

src/main/resources/hudson/plugins/emailext/MailAccount/config.groovy
Faheem12005: I had it set as the default account because I wasn't sure how the UI would work for additional accounts, and the Mailer plugin followed a similar layout. It was more of a preference, though, so I could change it if that would be better

Faheem12005: I tested it, works as expected

src/main/java/hudson/plugins/emailext/MailAccount.java
```diff
@@ -136,6 +150,81 @@ public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item item, @QueryPa
                     .includeCurrentValue(credentialsId);
         }
 
+        @SuppressWarnings("lgtm[jenkins/csrf]")
```
daniel-beck, line 153: This is an actual vulnerability. Sending an email is what's considered a "side effect".

@slide Please do not release the plugin in this state.

slide: The GSoC branch has not been merged into main, so there is no problem there.

daniel-beck: https://github.com/jenkinsci/email-ext-plugin/pull/1604

